PR #1, opened 2019-04-04
implemented task and tests

ivadimko:


src/arrayMethodSort.test.js
```diff
@@ -0,0 +1,52 @@
...
+const applyCustomSort = require('./arrayMethodSort');
+
+applyCustomSort();
+
+test('sort2 is added to [].__proto__', () => {
+  expect([].sort2)
+    .toBeInstanceOf(Function);
+});
+
+test('sort return a link to the original array', () => {
```
roman-mate: suggestion: sort returns original array.

src/arrayMethodSort.test.js
```diff
@@ -0,0 +1,52 @@
...
+  const result = source.sort2();
+  expect(source)
+    .toBe(result);
+});
+
+test('the original array is sorted', () => {
+  const source = [3, 12, 2, 11];
+  source.sort2();
+  expect(source)
+    .toEqual([11, 12, 2, 3]);
```
roman-mate: this is sorted as string. WAI?

src/arrayMethodSort.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+test('sort return a link to the original array', () => {
+  const source = [3, 12, 2, 11];
+  const result = source.sort2();
+  expect(source)
+    .toBe(result);
+});
+
+test('the original array is sorted', () => {
+  const source = [3, 12, 2, 11];
```
roman-mate: this array is not really sorted.

src/arrayMethodSort.test.js
```diff
@@ -0,0 +1,52 @@
...
+  source.sort2();
+  expect(source)
+    .toEqual([11, 12, 2, 3]);
+});
+
+test('sort numbers: [3, 12, 2, 11].sort2()', () => {
+  expect([3, 12, 2, 11].sort2())
+    .toEqual([11, 12, 2, 3]);
```
roman-mate: 12 is more than 11.

roman-mate: maybe update test description to smth like: "numbers are sorted as strings by default"

src/arrayMethodSort.test.js
ivadimko: updated description to 'the array is sorted'

PR #3, opened 2019-05-19
solution

valeraonisko:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,46 @@
...
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
+    if (this.length === 0) {
+      return this;
+    }
+    let compareFunctionInt = compareFunction;
+    if (typeof compareFunction === 'undefined') {
+      if (['string', 'number'].includes(typeof this[0])) {
+        compareFunctionInt = function(item1, item2) {
```
YuriiHoliuk: just `compareFunctionInt  = strCompare `

src/arrayMethodSort.js
```diff
@@ -3,9 +3,46 @@
...
+          return strCompare(item1, item2);
+        };
+      } else {
+        return this;
+      }
+    }
     // write code here
+    let count = 1;
```
YuriiHoliuk: looks like it may be Boolean

YuriiHoliuk: Would be good to extract sort algorithm into  a separate function

PR #12, opened 2019-08-21
Solution

PastushenkoVik:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+
+  return 0;
+};
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const testFunction = compareFunction === undefined
+      ? unicodeSort
+      : compareFunction;
```
irenhh: `const testFunction = compareFunction || unicodeSort;`

src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
 /**
  * Implement method Sort
  */
+const unicodeSort = (a, b) => {
+  const strA = String(a);
+  const strB = String(b);
+
+  for (let i = 0; i < strA.length; i++) {
```
irenhh:
```
  for (let i = 0; i < strA.length; i++) {
    if (!strB[i] || strA[i] > strB[i]) {
      return 1;
    } else if (strA[i] < strB[i]) {
      return -1;
    }
  }
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const testFunction = compareFunction === undefined
+      ? unicodeSort
+      : compareFunction;
+
+    for (let j = 0; j < this.length - 1; j++) {
+      for (let i = 0; i < this.length - 1; i++) {
+        if (testFunction(this[i], this[i + 1]) > 0) {
+          const tmp = this[i];
```
irenhh: use full names

src/arrayMethodSort.js
irenhh: what about this comment?

PR #13, opened 2019-08-24
Solution

ivansafiulin:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,34 @@
...
+
+  return 0;
+};
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const sortFunction = (compareFunction === undefined)
+      ? compareTo
+      : compareFunction;
```
irenhh: `const sortFunction = compareFunction || compareTo;`

PR #14, opened 2019-08-25
solution

shklyarigor2015:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,24 @@
...
  */
+function compareTo(a, b) {
+  return a.toString() > b.toString() ? 1 : -1;
+}
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    // eslint-disable-next-line max-len
+    const sortFunction = compareFunction !== undefined ? compareFunction : compareTo;
```
irenhh: `const sortFunction = compareFunction || compareTo;`

irenhh: there is no reason to disables this rule, wrap the line if it's too long

PR #16, opened 2019-08-26
solution Khrystyna

khrystynaStyslo:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,27 @@
...
+
+function compareElem(a, b) {
+  return a.toString() > b.toString() ? 1 : -1;
+}
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const sort
+      = compareFunction !== undefined ? compareFunction : compareElem;
```
irenhh: the right way to wrap the line
```
const sort = compareFunction !== undefined 
  ? compareFunction 
  : compareElem;
```

irenhh: another way
```
const sort = compareFunction || compareElem;
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,27 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        if (sort(this[j], this[j + 1]) > 0) {
+          const swap = this[j];
```
irenhh: `cachedValue`

PR #17, opened 2019-08-26
solution Andrii

Rafler:


src/arrayMethodSort.js
```diff
@@ -1,12 +1,24 @@
...
-
 /**
  * Implement method Sort
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const sortFunction = compareFunction !== undefined
+      ? compareFunction
+      : (a, b) => (String(a) > String(b) ? 1 : -1);
```
irenhh: another way
```
const sortFunction = compareFunction 
  || ((a, b) => (String(a) > String(b) ? 1 : -1));
```

src/arrayMethodSort.js
```diff
@@ -1,12 +1,24 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (sortFunction(this[i], this[j]) > 0) {
+          const tmp = this[i];
```
irenhh: use full names

PR #18, opened 2019-08-26
fs_jul19 Alex Andrutskii solution

aleks-and:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let swappedElements = true;
```
irenhh: `isSwapped` is better

irenhh (changes requested):

src/arrayMethodSort.js
aleks-and: Fixed!

irenhh (approved):

PR #35, opened 2020-01-27
Kravchenko: solution sort

sania517:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,79 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let functionCompare = function(first, second) {
+      const firsrStr = first + '';
+      const secondStr = second + '';
```
irenhh: use `String(...)`

PR #102, opened 2020-08-11
Solution

mpone:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
...
+    let count = 0;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const previousValue = this[i - 1];
+        const currentValue = this[i];
+
+        if (compareFunction(this[i - 1], this[i]) > 0) {
```
irenhh, line 33: why don't you use here the variables you created above?

PR #104, opened 2020-08-11
Solution

AnastasiiaBortnichuk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const asString = (a, b) => {
```
irenhh, line 7: quite a strange name for a function
suggestion: `compareStrings`

PR #157, opened 2020-12-17
Solution

vorfolomeevav:


MarchD (approved):
good solution )

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
-    // write code here
+  const stringsCompare = (a, b) => {
+    const stringA = a.toString();
+    const stringB = b.toString();
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringB > stringA) {
+      return -1;
+    } else {
```
varseniuk, line 15:
```suggestion
    } 
```

No need for this 'else' because it will work anyway if no of the previous checks won't pass.

varseniuk (approved):
Good 👍 

One comment

PR #169, opened 2020-12-19
solution

dtsehelnyk:


sasha39612 (approved):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,36 @@
 /**
  * Implement method Sort
  */
+const charsCompare = (charA, charB) => {
+  if (('' + charA) > ('' + charB)) {
```
varseniuk, line 7: You may use .toString method here.
It's more understandable for reading.
But if you want - leave as it is

src/arrayMethodSort.js
```diff
@@ -3,9 +3,36 @@
...
+  }
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = charsCompare) {
+    let counter = 0;
+
+    while (true) {
```
varseniuk, line 16: Let's try not to use `while(true)`. It's never a good idea. Especially in such kind of tasks.
Let's try to change your loop. 
E.g. to 'do... while (counter > 0)' loop and annulate counter on the BEGINNING of each iteration and increase 'counter' in case of elements' swap.

varseniuk (changes requested):
The logic of the solution is fine, but try to get rid of the `while(true)`

varseniuk (approved):
Great 👍

PR #181, opened 2020-12-21
Solution

Layrovell:


SofiiaAsieieva (approved):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,32 @@
 /**
  * Implement method Sort
  */
+
+const compareString = (a, b) =>
+  String(a) > String(b) ? 1
+    : String(b) > String(a) ? -1
+      : 0;
```
varseniuk, line 10:
```suggestion
const compareString = (a, b) =>
  String(a) > String(b) 
    ? 1
    : String(b) > String(a) 
      ? -1
      : 0;
```

If you want to use the ternary operator here, use formatting like this

varseniuk (approved):
Nice job!
But fix your formatting (commented)

PR #265, opened 2021-04-26
Solution

igor-stasiv:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,33 @@
 /**
  * Implement method Sort
  */
+function compareStrings(a, b) {
+  const strA = a.toString();
+  const strB = b.toString();
+
+  return strA > strB ? 1 : strA < strB ? -1 : 0;
```
Burunduk003, line 10: I suppose if we have more than two occasions it better to use if else condition.

Burunduk003 (approved):
Nice solution. You could put basic compare function in main function. I know that in Misha's example he did so, but in the task, we have 'write code here' field, so I suppose its a good practice to write code in that field)

src/arrayMethodSort.js
```diff
@@ -5,7 +5,42 @@
...
+
+        if (strA > strB) {
+          return 1;
+        }
+
+        if (strA < strB) {
+          return -1;
+        }
+
+        return 0;
```
varseniuk, line 24:
```suggestion
        return strA > strB;
```
You can simplify your code. If `strA > strB`, the function will return `true`. Otherwise - `false`. 
`true` is always `greater than` 0, so it is very useful in your comparison `f (compare(prev, cur) > 0)`

varseniuk (approved):
Great job! 👍 
Left one suggestion

PR #266, opened 2021-04-27
Solution

Burunduk003:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,40 @@
...
+
+    if (!compareFunction) {
+      compare = function(a, b) {
+        if (('' + a) > ('' + b)) {
+          return 1;
+        } else if (('' + a) === ('' + b)) {
+          return 0;
+        } else {
+          return -1;
+        }
```
igor-stasiv, line 19: Try to avoid **else** construction when you can, it will make code easier to read
```suggestion
        if (('' + a) > ('' + b)) {
          return 1;
        } 
        
        if (('' + a) === ('' + b)) {
          return 0;
        } 
        
        return -1;
 
```

igor-stasiv (approved):
Great job! And maybe it will be more understandable if you will use .toString() instead of (' ' +). But it is just my opinion)

src/arrayMethodSort.js
```diff
@@ -6,6 +6,40 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
mykola-yarchuk, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = () => {
    return something;
  }) {
```

you can do something like this

src/arrayMethodSort.js
```diff
@@ -6,6 +6,40 @@
...
+        }
+      };
+    }
+
+    let previousValue;
+    let currentValue;
+    let counter;
+
+    do {
+      counter = 0;
```
mykola-yarchuk, line 28: I like to use `true` `false` but 0 is also okay

src/arrayMethodSort.js
```diff
@@ -6,6 +6,40 @@
...
+
+      for (let i = 1; i < this.length; i++) {
+        currentValue = this[i];
+        previousValue = this[i - 1];
```
mykola-yarchuk, line 32: you can create constants here, in this way they would be deleted from the memory on every iteration of the loop.

src/arrayMethodSort.js
```diff
@@ -6,6 +6,40 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    let compare = compareFunction;
+
+    if (!compareFunction) {
+      compare = function(a, b) {
+        if (('' + a) > ('' + b)) {
```
mykola-yarchuk, line 13:
```suggestion
  const aItem = String(a);
  const bItem = Stirng(b);
```

mykola-yarchuk: and use them, no need to do this every time
```
'' + a
```

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):
graet

PR #268, opened 2021-04-28
Solution

nazarmatsevych:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,34 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let callback = compareFunction;
+    let count;
+
+    if (!callback) {
+      callback = function(a, b) {
```
mykola-yarchuk, line 12: you can write this callback as default value for function's parameter

mykola-yarchuk (changes requested):
almost done

mykola-yarchuk (approved):
great

PR #272, opened 2021-04-29
Develop

SerhiiKolesnyk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = ((prevValue, currentValue) =>
```
victor-arutiunov, line 8: Nice solution, but you can put your default function down with if operator to make code more readable.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
...
+      prevValue.toString() > currentValue.toString())
+  ) {
+    let counter;
+
+    do {
+      counter = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const prevValue = this[i - 1];
+        const currentValue = this[i];
```
mykola-yarchuk, line 18: no need to keep both of them one would be enough

mykola-yarchuk (approved):
great

PR #275, opened 2021-04-30
develop

Andrey203:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,24 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => `${a}` > `${b}`) {
```
Ihor-Karpyn, line 7: I think in this case toString method more understandable

src/arrayMethodSort.js
```diff
@@ -4,8 +4,24 @@
...
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        if (compareFunction(this[i - 1], this[i]) > 0) {
+          const el = this[i - 1];
```
Ihor-Karpyn, line 15: el it's bad var name

Ihor-Karpyn (approved):
Good job, left couple comments

aliceblunt8 (approved):

PR #277, opened 2021-04-30
Solution v0.1

SkyPKWolf:


Allrightman:
Nice! 
I was confused with only 2 output (not 3) but yeah, true or false works the same way (in case of "sort")

Allrightman (approved):
Nice!
I was confused with only 2 output (not 3) but yeah, true or false works the same way (in case of "sort")

src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
...
+    compareFunction = (a, b) => String(a) > String(b)
+  ) {
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const current = this[i];
+        const prev = this[i - 1];
```
mykola-yarchuk, line 17: одного значення достатньо, не потрбіно сворювати 2 змінних

mykola-yarchuk (approved):
great

PR #278, opened 2021-04-30
solution

Allrightman:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
...
+    if (strA > strB) {
+      return 1;
+    } else if (strA === strB) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
+
+  [].__proto__.sort2 = function(compareFunction = standartCompare) {
```
Ihor-Karpyn, line 20: I think that using external variables in a function is a bad practice. Also you can simplify this.
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => a.toString() > b.toString() {
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
...
+    let prev = this[0];
+    let curr = this[1];
```
Ihor-Karpyn, line 22: redundant

src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
...
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        prev = this[i - 1];
+        curr = this[i];
```
Ihor-Karpyn, line 30:
```suggestion
        const prev = this[i - 1];
        const curr = this[i];
```

Ihor-Karpyn (changes requested):

src/arrayMethodSort.js
Allrightman: oh really, thanks!

Ihor-Karpyn (approved):

aliceblunt8 (approved):

PR #283, opened 2021-05-03
solution

RuslanaPodshyvalova:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+          return -1;
+        }
+
+        if (String(a) > String(b)) {
+          return 1;
+        }
+
+        return 0;
+      };
+    }
```
Ihor-Karpyn, line 22:
```suggestion
[].__proto__.sort2 = function(compareFunction = (a, b) => a.toString() > b.toString()) {
```
Enough

src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+
+    let value;
+
+    for (let i = this.length - 1; i > 0; i--) {
```
Ihor-Karpyn, line 26: use do while, your solution very difficult. When you don't need make change you do extra iterations

Ihor-Karpyn (changes requested):

aionova (changes requested):
Let's try to do it without loop inside of another loop

src/arrayMethodSort.js
```diff
@@ -11,17 +11,20 @@ function applyCustomSort() {
       comparator = (a, b) => a.toString() > b.toString();
     }
 
-    let tmpValue;
+    let tmpValue, isExchange;
```
varseniuk, line 14: You can initialize the variable `tmpValue` directly in your  `for` loop. It's needed only there and on each iteration, it has different values. It is not needed outside the loop

varseniuk (approved):
Great 🚀

PR #286, opened 2021-05-04
solution

IvanKolodiazhnyi:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,28 @@
...
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+const noArguments = (a, b) => a.toString() > b.toString();
+
```
Ihor-Karpyn, line 30: I think that using external variables in a function is a bad practice.
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => a.toString() > b.toString()) {
    let count = 0;

    do {
      count = 0;

      for (let i = 1; i < this.length; i++) {
        const prev = this[i - 1];
        const current = this[i];

        if (compareFunction(prev, current) > 0) {
          this[i - 1] = current;
          this[i] = prev;
          count++;
        }
      }
    } while (count > 0);

    return this;
  };
}



```

Ihor-Karpyn (approved):

src/arrayMethodSort.js
```diff
@@ -4,9 +4,28 @@
...
+          count++;
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+const noArguments = (a, b) => a.toString() > b.toString();
```
VinnieJ-2k20, line 29: The safest way is to use function declaration here.
And better call it something like `defaultSort`

VinnieJ-2k20 (changes requested):
Almost done.
One fix required.

VinnieJ-2k20 (approved):
Good job 👍

PR #287, opened 2021-05-04
solution

vitalya2626:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,24 @@
...
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    if (String(a) > String(b)) {
+      return 1;
+    } else if (String(a) < String(b)) {
+      return -1;
+    } else {
+      return 0;
+    }
+  }) {
```
Ihor-Karpyn, line 15:
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => {
    if (String(a) > String(b)) a.toString() > b.toString()) {
```
It's enough

src/arrayMethodSort.js
```diff
@@ -4,8 +4,24 @@
...
+    for (let i = 0; i < this.length; i++) {
```
Ihor-Karpyn, line 16: use do while, your solution very difficult. When you don't need make change you do extra iterations

Ihor-Karpyn (changes requested):

aionova (changes requested):
Let's follow Ihor's advice. Get rid of loop inside of loop (you can use do while instead).

aionova (approved):
Looks great!

PR #293, opened 2021-05-05
js_array-method-sort

Bogdan-Kotsupey:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareFunction = (a, b) => {
```
victor-arutiunov, line 7: I think that it would be better if you move default callback into sort method.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
...
+    const strA = String(a);
+    const strB = String(b);
```
victor-arutiunov, line 9: You can transfer "String(x)" into if operator, without assign.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
...
+
+    if (strA > strB) {
+      return 1;
+    } else {
```
VinnieJ-2k20, line 13: If you use return inside `if` block, writing `else` is useful. This is the correct structure:
```
if (condition1) {
  return x;
};

if (condition2) {
  return y;
};

return z;
```

VinnieJ-2k20 (changes requested):
Solution looks good, just need to fixe some syntax.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,37 @@
...
+
+  [].__proto__.sort2 = function(callback = compareFunction) {
+    let count = 0;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const firstElem = this[i - 1];
+        const secondElem = this[i];
```
mykola-yarchuk, line 27: you can save only one element

mykola-yarchuk (approved):
great

PR #342, opened 2021-08-07
done

SashaLifashkin:


ivanchenko-kirill (approved):
Отличная работа 🔥

src/arrayMethodSort.js
```diff
@@ -5,7 +5,43 @@
...
+
+    if (!compareFunction) {
+      const defaultCompare = (a, b) => {
+        if (String(a) > String(b)) {
+          return 1;
+        } else if (String(a) < String(b)) {
+          return -1;
+        } else {
+          return 1;
+        }
```
herkoss, line 21:
```suggestion
        if (String(a) > String(b)) {
          return 1;
        }
        
        if (String(a) < String(b)) {
          return -1;
        }
        
        return 0;
```

herkoss: If an if block always executes a return statement, the subsequent else block is unnecessary. A return in an else if block following an if block that contains a return can be separated into multiple if blocks

More details here
mate-academy.github.io/style-guides/javascript.html#blocks--no-else-return

herkoss (changes requested):
Well done, but it needs some refactoring for better readability)
Good luck!🤞🍀

herkoss (approved):
Great!👍😌

PR #347, opened 2021-08-09
Solution

Anna-Tkachenko:


alex-glotov67 (approved):
🔥

src/arrayMethodSort.js
```diff
@@ -4,9 +4,29 @@
...
+          changesCounter++;
+        }
+      }
+    } while (changesCounter > 0);
+
+    return this;
   };
+
+  function compareAsStrings(a, b) {
+    return (String(a) > String(b)) ? 1 : -1;
```
VinnieJ-2k20, line 28: Sort callback should also return 0 if `a` and `b` are the same.

VinnieJ-2k20 (changes requested):
Nice solution. It works with current tests.
But you need to be as close to original sort method as possible.

src/arrayMethodSort.js
```diff
@@ -4,9 +4,42 @@
...
+      return 1;
+    }
+
+    if (x < y) {
+      return -1;
+    }
+
+    if (x === y) {
+      return 0;
+    }
```
VinnieJ-2k20, line 41: Final `if` is redundant.
```suggestion
    if (x < y) {
      return -1;
    }

    return 0;
```

VinnieJ-2k20 (approved):
Good job 👍

Clonvhs (commented):
Great one

PR #348, opened 2021-08-10
solution

VeronikaLushchik:


SashaLifashkin:
Хорошая работа!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,30 @@
...
+          check = true;
+        }
+      }
+    } while (check);
+
+    return this;
   };
+
+  function compare(a, b) {
+    return (String(a) > String(b)) ? 1 : -1;
```
herkoss, line 29: What if strings are equal?

herkoss (changes requested):
Not bad, but there is one little mistake)
Let's fix it)

src/arrayMethodSort.js
```diff
@@ -4,9 +4,30 @@
...
+          check = true;
+        }
+      }
+    } while (check);
+
+    return this;
   };
+
+  function compare(a, b) {
+    return (String(a) === String(b)) ? 0 : (String(a) > String(b) ? 1 : -1);
```
herkoss, line 29: Using a ternary operator in the ternary operator is bad practice. Use if/else instead

herkoss (changes requested):
Well done, but there is one little mistake.
Let's fix it!)
Good luck🤞🍀

herkoss (approved):
Awesome!👏😎

PR #349, opened 2021-08-10
sort

nastyyvolk:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,40 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringB > stringA) {
+    return -1;
+  } else {
+    return 0;
+  }
```
illiarerun, line 13: I think that else keyword in second condition and last block are not much needed here, cause you return values if condition is true (not sure though)
```suggestion
  if (stringA > stringB) {
    return 1;
  }
  
  if (stringB > stringA) {
    return -1;
  }
  
  return 0;
```

illiarerun (approved):
Great job 🔥

src/arrayMethodSort.js
VinnieJ-2k20: Yep, Illia is correct. If you have `return` inside `if`, the following `else` is redundant.

VinnieJ-2k20 (changes requested):
Almost done.
Need to fix `if` structure.

VinnieJ-2k20 (approved):
Good job 👍

Clonvhs (commented):
Good one

PR #352, opened 2021-08-11
sorted

ihor-jpeg:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,25 @@
...
+        }
+      }
+    }
+
+    return this;
+  };
+
+  const compareAsStrings = (itemOne, itemTwo) => {
+    return String(itemOne) > String(itemTwo) ? 1 : -1;
   };
```
Anna-Tkachenko, line 26: I had the same function, and my solution was rejected because we need to be as close to original sort method as possible (this method should return 0 if itemOne and itemTwo are the same)

src/arrayMethodSort.js
```diff
@@ -4,8 +4,25 @@
...
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
+    let counter = 1;
+
+    while (counter) {
+      counter = 0;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        if (compareFunction(this[i], this[i + 1]) > 0) {
+          [this[i], this[i + 1]] = [this[i + 1], this[i]];
```
Anna-Tkachenko, line 15: 👍

Anna-Tkachenko (changes requested):

src/arrayMethodSort.js
ihor-jpeg: thank you!
but in that case i think it should be specified. there is literally no situation why we would need to know if two numbers are equal. this is redundant to the sorting process.
i'm inclined to keep it this way unless someone can show me that it is actually wrong and why

herkoss: You can read about array.sort [here](https://262.ecma-international.org/10.0/#sec-sortcompare)

Еhe interpretation of sorting may differ in different browsers, so you need an implementation that works the same everywhere

herkoss (changes requested):
Good job!👍
But there is still work to be done)
Good luck🤞

herkoss (approved):
Well done!👏✅

Clonvhs (commented):
Great one

PR #354, opened 2021-08-11
solution

dima-tkachenko:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
VinnieJ-2k20, line 16: We don't need `else` if we have `return` inside `if` structure.
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA === stringB) {
    return 0;
  }
  
  return -1;
```

VinnieJ-2k20 (changes requested):
Solution is good, but let's fix code style.

VinnieJ-2k20 (approved):
Good job 👍

PR #357, opened 2021-08-12
solution

DmytroMykoliv:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
...
+
+    return this;
   };
 }
 
+const compareStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
```
dima-tkachenko, line 29: if (stringA > stringB) {
    return 1;
  } ;

if (stringA === stringB) {
    return 0;
  };

return -1;

виглядає краще)

dima-tkachenko:
цікаве рішення!:fire:

src/arrayMethodSort.js
DmytroMykoliv: Дякую, згідний ))

src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
...
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  }
```
herkoss, line 33:
```suggestion
  }
  
  if (stringA === stringB) {
    return 0;
  }
```

herkoss: If an if block always executes a return statement, the subsequent else block is unnecessary. A return in an else if block following an if block that contains a return can be separated into multiple if blocks

More details here
mate-academy.github.io/style-guides/javascript.html#blocks--no-else-return

herkoss (changes requested):
Well done!👏
But it needs some refactoring)

src/arrayMethodSort.js
DmytroMykoliv: Thanks, i missed it )
Done

VinnieJ-2k20 (approved):
Good job 👍

PR #358, opened 2021-08-12
solved

lera-yaskevych:


DmytroMykoliv (approved):
Looks good )

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
+  }
```
VinnieJ-2k20, line 18: When we write `return` inside `if` - `else` is redundant.
```suggestion
    if (stringA > stringB) {
      return 1;
    } 
    if (stringA === stringB) {
      return 0;
    }
    
    return -1;
  }
```

VinnieJ-2k20 (changes requested):
Solution is correct, but let's fix `if` structure.
And make sure to pay attention to it in future tasks.

src/arrayMethodSort.js
lera-yaskevych: I somehow missed that. That's obvious. Thanks!

VinnieJ-2k20 (approved):
Good job 👍

PR #366, opened 2021-08-13
sortMethod

zikozakazabiyaka:


evgen1sider (approved):
Great!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringA = a.toString();
+  const stringB = b.toString();
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringB > stringA) {
+    return -1;
+  } else {
+    return 0;
+  }
```
VinnieJ-2k20, line 39: If you write `return` inside `if` , `else` is redundant.
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringB > stringA) {
    return -1;
  }
  
  return 0;
```

VinnieJ-2k20 (changes requested):
Almost done, just fix `if` structure.

VinnieJ-2k20 (approved):
Good job 👍

PR #368, opened 2021-08-13
Sort

andrii-hryhoruk:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
 /**
  * Implement method Sort
  */
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
```
sasha-morozov, line 10: Else is redundant. You can return "-1" without using "else". Apart from this, you have to add paddings between your if statements.

sasha-morozov (approved):
Nice solution!)
Left one suggestion above

Denchepornyuk (approved):

PR #369, opened 2021-08-13
Develop

Vova-Stoliar:


andrii-hryhoruk:
Nice solution)

andrii-hryhoruk (approved):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,44 @@
...
+  switch (true) {
+    case aToString > bToString:
+      return 1;
+
+    case aToString === bToString:
+      return 0;
+
+    default:
+      return -1;
+  }
```
VinnieJ-2k20, line 20: This use of `switch/case` is valid, but not very appropriate. Rewrite using `if/return` structure.

VinnieJ-2k20 (changes requested):
Solution is correct, but let's fix some code style.

src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+
+const compareAsString = (a, b) => {
+  const aToString = String(a);
+  const bToString = String(b);
+
+  if (aToString > bToString) {
+    return 1;
+  } else if (aToString === bToString) {
+    return 0;
+  }
```
VinnieJ-2k20, line 15: If you write `return` inside `if` , `else` is redundant.
```suggestion
  if (aToString > bToString) {
    return 1;
  } 
  
  if (aToString === bToString) {
    return 0;
  }
```

VinnieJ-2k20 (changes requested):
Final fix to `if` structure.

VinnieJ-2k20 (approved):

PR #370, opened 2021-08-13
solution

derelicttt:


MaksymBeliever (approved):
Good job 👍

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
VinnieJ-2k20, line 16: If `return` is written inside `if` , `else` is redundant.
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA === stringB) {
    return 0;
  } 
  
  return -1;
```

VinnieJ-2k20 (changes requested):
Almost done, just rewrite `if` structure.

VinnieJ-2k20 (approved):
Good job 👍

PR #371, opened 2021-08-13
Solution

MaksymBeliever:


set1one (approved):
Good

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return null;
+    } else {
+      return -1;
+    }
```
varseniuk, line 17:
```suggestion
    if (stringA > stringB) {
      return 1;
    } else {
      return -1;
    }
```
You need to return `1` only if a > b

varseniuk (approved):
Good, but pay attention to my comment

PR #372, opened 2021-08-13
Solution

poliandiya:


LazurNiko (approved):
nice work!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
 }
 
+function compareStrings(a, b) {
+  if (String(a) > String(b)) {
+    return 1;
+  }
+
+  if (String(a) < String(b)) {
+    return -1;
+  }
```
VinnieJ-2k20, line 37: You do same type conversion twice. Better to use variables and do it only once.

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment on how to simplify your code.

PR #373, opened 2021-08-13
js_array-method-sort

RomanFedusevych:


LazurNiko (approved):
Great!

src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+
+const compareAsSrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  }
```
VinnieJ-2k20, line 15: When `return` is written inside `if` , `else` is redundant.
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA === stringB) {
    return 0;
  }
```

VinnieJ-2k20 (changes requested):
Almost done, let's rewrite `if` structure.

VinnieJ-2k20 (approved):
Good job 👍

PR #374, opened 2021-08-15
solution

LazurNiko:


Anna-Tkachenko (changes requested):
Nice Going👍
Left some comments about code style

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
 
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  }
```
VinnieJ-2k20, line 37: If we write `return` inside `if` , `else` is redundant.

VinnieJ-2k20 (changes requested):
Almost done, let's fix your code style.

src/arrayMethodSort.js
```diff
@@ -4,9 +4,37 @@
...
   };
 }
 
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  }
```
VinnieJ-2k20, line 35: `else` is redundant, but condition still needs to be checked.
```suggestion
  if (stringA > stringB) {
    return 1;
  }
  
  if (stringA === stringB) {
    return 0;
  }
```

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):

PR #375, opened 2021-08-15
Solution

Illia-Boiko:


set1one (approved):
Great!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,28 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = stringSort) {
+    let marker = 1;
+
+    while (marker !== 0) {
+      marker = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        if (compareFunction(this[i - 1], this[i]) >= 1) {
```
varseniuk, line 14:
```suggestion
        if (compareFunction(this[i - 1], this[i]) > 0) {
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,28 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = stringSort) {
+    let marker = 1;
+
+    while (marker !== 0) {
+      marker = 0;
```
varseniuk, line 11: Use `do.. while` instead of assigning `marker = 1` and check `while (marker > 0)`

varseniuk (changes requested):
Logic is good, but let's make it even better

varseniuk (approved):

PR #376, opened 2021-08-15
solution by set1one

set1one:


Clonvhs (approved):
Good one

src/arrayMethodSort.js
```diff
@@ -4,8 +4,32 @@
...
+    return this;
+  };
+
+  const initialCompare = (a, b) => {
+    const firstItem = String(a);
+    const secondItem = String(b);
+
+    if (firstItem === secondItem) {
+      return 0;
+    }
```
varseniuk, line 30:
```suggestion
```
In fact, there's no need in this here

src/arrayMethodSort.js
```diff
@@ -4,8 +4,32 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = initialCompare) {
+    let count = 1;
+
+    while (count) {
+      count = 0;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        if (compareFunction(this[i], this[i + 1]) > 0) {
```
varseniuk, line 14:
```suggestion
        if (compareFunction(this[i], this[i + 1]) > 0) {
```
For better readability it would be better to create variables for `this[i]` and `this[i + 1]`. Like `next` and `current`
```

varseniuk (approved):
Good.
Pay attention to my comments

PR #380, opened 2021-08-18
Solution

Ingvvvar:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,37 @@
...
  */
+function compareAsStrings(a, b) {
+  const stringA = `${a}`;
+  const stringB = `${b}`;
+
+  return stringA > stringB
+    ? 1
+    : stringA < stringB
+      ? -1
+      : 0;
```
illiarerun, line 14: Paul suggested me to remove nested ternaries, cause they make code a bit complicated. So I think you should do it, too)

illiarerun (approved):
Good job 🔥 But remove these nested ternary thing and it will be great 😄

src/arrayMethodSort.js
varseniuk: Illia is right

varseniuk (approved):
Good job, but pay attention to the Illia's comment

src/arrayMethodSort.js
Ingvvvar: Ilya is always right!

PR #392, opened 2021-09-20
solution

Daerniss:


CouldSweeT (approved):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,40 @@
...
+          if (compareFunction(this[i], this[j]) > 0) {
+            const temp = this[j];
+
+            this[j] = this[i];
+            this[i] = temp;
+          } else {
+            continue;
+          }
+        }
+      }
```
Anna-Tkachenko, line 38: Rewrite it using cycle 'while' as in original sort method implementation, don't use nested loops

src/arrayMethodSort.js
```diff
@@ -5,7 +5,40 @@
...
+        let smallestIndex = 0;
+
+        for (let j = 0; j < copy.length - 1; j++) {
+          if (copy[j].toString() > copy[j + 1].toString()) {
+            smallestIndex = j + 1;
+          }
+        }
+        this[i] = copy.splice(smallestIndex, 1)[0];
+        smallestIndex = 0;
+      }
```
Anna-Tkachenko, line 25: Same here, try to rewrite it without nested loops

Anna-Tkachenko (changes requested):

PR #399, opened 2021-09-22
solution

CouldSweeT:
git commit -am solution


Anna-Tkachenko (changes requested):
Try to rewrite it using cycle 'while' as in the original sort method implementation, don't use nested loops

Anna-Tkachenko (approved):

PR #408, opened 2021-10-14
Solution

MariiaObraztsova:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultSortedAsStrings) {
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const prev = this[i - 1];
```
Igor-MateAcademy, line 14: Don't slice variables name)

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+          this[i] = prev;
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
+
+  function defaultSortedAsStrings(element1, element2) {
+    const string1 = element1.toString();
```
Igor-MateAcademy, line 30: On my opinion, use structure's name in variable's name is bad practice. Once time, Vova reject my task, because i was used a structure's name in my variable's name)

Igor-MateAcademy (approved):

src/arrayMethodSort.js
solaryasha: Agree with Igor, let's change change names for proper one's

PR #427, opened 2021-11-11
Implement my own custom sort method using bubble sort

dmtrhrytsak:


src/arrayMethodSort.js
```diff
@@ -5,8 +5,37 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const compare = compareFunction || compareAsStrings;
```
illiarerun, line 8: You can just set a default value in function params

src/arrayMethodSort.js
```diff
@@ -5,8 +5,37 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      const unsortedElement = this[i];
```
illiarerun, line 11: be more specific
```suggestion
      const firstUnsortedElement = this[i];
```

illiarerun (changes requested):
Good job 🔥  Let's few things:

src/arrayMethodSort.js
dmtrhrytsak: Oh, stupid me :)

src/arrayMethodSort.js
dmtrhrytsak: I'll stick to my naming, there are no second or third unsortedElement down the line

illiarerun: You got me on this one)))

illiarerun: image

illiarerun (approved):
Good job 🔥

PR #428, opened 2021-11-12
Sort method

dariia-romanova:


JamesSunderlanddd (approved):
Good Job!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+
+    do {
+      countChanges = 0;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        if (compareFunction(this[i], this[i + 1]) > 0) {
+          const temp = this[i];
+
+          this[i] = this[i + 1];
+          this[i + 1] = temp;
```
ihor-jpeg, line 18: as a useful thing to know, think (or google) how we can swap two values without a third variable. another useful thing would be figuring out how it works :)

ihor-jpeg (changes requested):
The code works well. But there are a few things to work on.
When I run `[112, 44].sort2()`, your sorting method doesn't work. It returns [112, 44] instead of [44, 112]

ihor-jpeg (approved):
oh, I see. The default sort as strings was intended. My bad

src/arrayMethodSort.js
dariia-romanova: Changed to destructuring assignment)

PR #437, opened 2021-11-17
Solution

olya-shyian:


volodymyr-soltys97 (approved):
GJ!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,22 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    return String(a) > String(b);
+  }) {
```
aionova, line 9: let's write default function separately. It will be easier to read your code then

aionova (approved):

PR #441, opened 2021-11-17
Solution

msaikun:


volodymyr-soltys97 (approved):
Very well! 👍

src/arrayMethodSort.js
```diff
@@ -4,9 +4,26 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => a.toString() > b.toString()) {
```
aionova, line 8: default compare function is better to be written separately. It will be easier to read your code then

aionova (approved):

PR #444, opened 2021-11-18
Solution

4qwerty:


mariia-bila (approved):
Nice work!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA < stringB) {
+      return -1;
+    }
+
+    if (stringA > stringB) {
+      return 1;
+    }
```
aionova, line 17: code will be easier to read if you write default compare function separately and use it here

aionova (approved):
GJ!

PR #446, opened 2021-11-18
solution

mariia-bila:


IrynaHerasymchuk (approved):

src/arrayMethodSort.js
```diff
@@ -5,8 +5,43 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let result = 0;
+    let first, next;
+    let count = 0;
+
+    if (compareFunction === undefined) {
```
aionova, line 12: let's give a default value to compare function in parameters and get rid of 'if else' here

aionova (changes requested):

PR #473, opened 2021-12-22
solution

Basil-Panasiuk:


ArturErmolenko (approved):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+    return 0;
+  }
+
+  return -1;
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
```
Elena-brv, line 21: good job 🔥 

also, you can apply by default such function and it will work the same
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b) ) {
```

Elena-brv (approved):

PR #499, opened 2022-01-28
solution

013g:


KostyaKovbel (approved):
GG WP!!!

src/arrayMethodSort.js
```diff
@@ -3,10 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA < stringB) {
+    return -1;
+  } else {
+    return 0;
+  }
```
Anna-Tkachenko, line 16: Last else is redundant
```suggestion
  }
  
  return 0;
```

Anna-Tkachenko (approved):

PR #501, opened 2022-01-29
sol

KostyaKovbel:


pogrebnyakp (approved):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,44 @@
...
+    return 1;
+  }
+
+  if (stringOne === stringTwo) {
+    return 0;
+  }
+
+  if (stringOne < stringTwo) {
+    return -1;
+  }
```
Anna-Tkachenko, line 21: One if is redundant here, you can return one value by default

Anna-Tkachenko (approved):

PR #524, opened 2022-02-18
solution

RomanKushyk:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+const compareAsString = (a, b) => {
+  if (`${a}` > `${b}`) {
```
krolich500grn, line 30: interesting idea, but i think it have less readability that default String(a)

krolich500grn (approved):
good job! 👍

src/arrayMethodSort.js
krolich500grn: @solaryasha так апрувнув ж)

PR #580, opened 2022-05-02
Task complete

rHashTag:
d(o0,)b

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -3,9 +3,44 @@
 /**
  * Implement method Sort
  */
+function campareAsStrings(first, second) {
+  let a = first;
+  let b = second;
+
+  a = String(a);
+  b = String(b);
+
+  if (a > b) {
+    return 1;
+  }
+
+  if (a < b) {
+    return -1;
+  }
+
+  return 0;
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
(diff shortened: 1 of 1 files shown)


src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+  }
+
+  return 0;
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = campareAsStrings) {
+    let count;
```
huberciak, line 23: I dont know why but I got confused by this variable name.If I understand your code, count is  variable checking if some strings changed indexes. Also you are checking just >0 statement so maybe use boolean type. 
Its just my thoughts.

huberciak (approved):
Nice code, only my personal thought added. Nice job!

src/arrayMethodSort.js
VinnieJ-2k20: That's fine, imho.

src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
 /**
  * Implement method Sort
  */
+function campareAsStrings(firstElement, secondElement) {
```
VinnieJ-2k20, line 6:
```suggestion
function compareAsStrings(firstElement, secondElement) {
```

VinnieJ-2k20 (approved):
Good job 👍 
Check some spelling.

PR #581, opened 2022-05-03
solution

efdalyalcin:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,39 @@
...
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        if (compareFunction(this[i - 1], this[i]) > 0) {
+          const placeholder = this[i];
+
+          this[i] = this[i - 1];
+          this[i - 1] = placeholder;
```
rHashTag, line 33: This sections gets a bit confusing, `this[i]` and `this[i - 1]` are used more than once, and the variable `const placeholder = this[i]` is also created. For readability, perhaps using variables for both  `this[i]` and `this[i - 1]` in this loop would be a better idea.

rHashTag (changes requested):
Great job, code looks amazing!

I've added one suggestion.

src/arrayMethodSort.js
efdalyalcin: created the variables

rHashTag (approved):
LGTM!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const createdSortFunc = (a, b) => {
```
VinnieJ-2k20, line 7: Choose more appropriate name for this function.

VinnieJ-2k20 (changes requested):
Almost done. One small edit left.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const comparisonOfStrings = (a, b) => {
```
VinnieJ-2k20, line 7: Try again. Function nave should start with a verb.

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #591, opened 2022-05-06
solution

huberciak:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,40 @@
 /**
  * Implement method Sort
  */
+function compare(A, B) {
+  const a = '' + A;
+  const b = '' + B;
```
rHashTag, line 8: I noted that you've changed the function name as well as the arguments. I would believe that keeping the function name would be better as well as the argument names.

rHashTag (changes requested):
Good job on the task. You took a different approach which quite interesting.

I've aded 1 suggestion.

meforoma (approved):
great implementation.
one non-blocking comment is to avoid capitalized variables
- usually capitalized ones are either Global variables or React Components

PR #597, opened 2022-05-08
Develop

PRO-GRAM-MER:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,44 @@
...
 
-/**
- * Implement method Sort
- */
+function compareFormstring(first, second) {
+  let a = first;
+  let b = second;
+
+  a = String(a);
+  b = String(b);
```
rHashTag, line 8: You can actually combine these 4 lines into 2.

```suggestion
  const a = String(first);
  const b = String(second);
```

Rashaad-Ebrahim (changes requested):
Good job bud! I've added 1 suggestions just to optimize the code a bit.

meforoma (approved):

PR #598, opened 2022-05-09
sort

elkhayate:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  function sortString(a, b) {
+    const strA = String(a);
+    const strB = String(b);
+
+    if (strA > strB) {
+      return 1;
+    } else if (strA === strB) {
+      return 0;
```
MabuzaM, line 14: I think it can still work fine without checking for this condition.

MabuzaM (changes requested):
Good job. Please check the comment below.

src/arrayMethodSort.js
elkhayate: ur right, thank you.

meforoma (approved):

PR #631, opened 2022-06-13
Solution

stanislav-topikha:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,36 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
+function defaultSort(a, b) {
+  if (`${a}` > `${b}`) {
```
Pavlo000, line 4: why do you need to use interpolation here?

src/arrayMethodSort.js
```diff
@@ -1,11 +1,36 @@
...
+    return 1;
+  }
+
+  if (`${a}` === `${b}`) {
```
Pavlo000, line 8: Here also

Pavlo000 (approved):

src/arrayMethodSort.js
stanislav-topikha: It has to be strings

Olexandr6, line 4:
```suggestion
  if (String(a) > String(b)) {
```

src/arrayMethodSort.js
Olexandr6: The same here

Olexandr6 (approved):

PR #634, opened 2022-06-13
Solution method Sort

ghost:


senile-pitcher (approved):
👍

src/arrayMethodSort.js
```diff
@@ -4,9 +4,30 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length; j++) {
+        if (compareFunction(this[j], this[i]) > 0) {
+          const temporaryValue = this[i];
+
+          this[i] = this[j];
+          this[j] = temporaryValue;
+        }
+      }
+    }
```
Olexandr6, line 27: How many iterations will be if the array is already sorted?

src/arrayMethodSort.js
```diff
@@ -4,9 +4,30 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  function compareString(a, b) {
+    if (String(a) > String(b)) {
+      return 1;
+    } else if (String(a) < String(b)) {
+      return -1;
+    } else {
+      return 0;
+    }
```
Olexandr6, line 14: You used String() constructor 4 times, let's use only 2 times

G-YaSha (changes requested):
Your solution is good👍, but can be improved

src/arrayMethodSort.js
Olexandr6: Read about a flag in bubble sort

jstmpelowycz (approved):

PR #639, opened 2022-06-14
Develop

ostapfriend:


vsupruniuk (approved):
Nice👍

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+      }
+    } while (swap);
+
+    return this;
   };
 }
 
+const compareStrings = (a, b) => {
+  const strA = '' + a;
+  const strB = '' + b;
```
Olexandr6, line 31:
```suggestion
  const strA = String(a);
  const strB = String(b);
```

Olexandr6: Better to use String constructor

Olexandr6 (approved):

PR #656, opened 2022-06-18
first commit

pavel-gutsal:
first commit

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -1,11 +1,37 @@
 'use strict';
-
 /**
  * Implement method Sort
  */
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i; j < this.length; j++) {
+        const changeHandler = (a, b) => {
+          const temp = this[a];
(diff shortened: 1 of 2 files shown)


src/arrayMethodSort.js
```diff
@@ -1,11 +1,37 @@
 'use strict';
-
 /**
  * Implement method Sort
  */
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
vkryvytskyy, line 7: add here default value for the compareFunction
```suggestion
  [].__proto__.sort2 = function(compareFunction = yourFunction) {
```

src/arrayMethodSort.js
```diff
@@ -1,11 +1,37 @@
...
     // write code here
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i; j < this.length; j++) {
```
vkryvytskyy, line 10: consider to use `do..while` approach with nested `for`

vkryvytskyy (changes requested):

pavel-gutsal:
do while loop implemented with nested  for loop;
defualt sort  callback function was written

vkryvytskyy (approved):

PR #660, opened 2022-06-22
Solution

averstiuk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
jstmpelowycz, line 17: you don't need else-if/else statements

jstmpelowycz (changes requested):

mykola-koval (approved):

PR #670, opened 2022-07-04
solution

oleksandr-kotliarov:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const defaulCompare = (a, b) => {
```
VETALEK, line 7: you may use `localCompare` here

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    } else {
+      return -1;
+    }
+  };
+
+  [].__proto__.sort2 = function(compareFunction = defaulCompare) {
+    let count;
+
+    do {
+      count = 0;
```
VETALEK, line 24: `count` is not descriptive enough

VETALEK (approved):
👍

PR #677, opened 2022-07-07
Solution

Sergii2705:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
...
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (itemA, itemB) =>
+    String(itemA) > String(itemB) ? 1 : 0) {
+    for (let loop = 1; loop < this.length; loop++) {
+      for (let item = 0; item < this.length - loop; item++) {
+        if (compareFunction(this[item], this[item + 1]) > 0) {
+          [this[item], this[item + 1]] = [this[item + 1], this[item]];
+        }
+      }
+    }
```
VETALEK, line 15: you need to check if there is any need to continue sort on each iteration, cause even if we have sorted array - you function will still loop through it many times

VETALEK (changes requested):
Good start 👍 
commented

src/arrayMethodSort.js
Sergii2705: thanks, correct remark, I thought about this when I wrote the code, I even have ideas for implementing this, it's just that nothing is written about it in the condition. No problem, I'll do it now.

Ihor-Karpyn (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,26 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (itemA, itemB) =>
+    String(itemA) > String(itemB) ? 1 : 0) {
+    let isSorted;
+
+    for (let loop = 1; loop < this.length; loop++) {
```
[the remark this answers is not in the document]
Ihor-Karpyn: Good point, I overlooked that. So u should change the variable name.

Ihor-Karpyn (approved):

PR #684, opened 2022-07-12
Oleksandr-Pustovit

Oleksandr-Pustovit:


dashasm (approved):

src/arrayMethodSort.js
```diff
@@ -5,8 +5,46 @@
...
+
+function defaultCompareFunction(a, b) {
+  const first = '' + a;
+  const second = '' + b;
+
+  switch (true) {
+    case first < second: return -1;
+    case first > second: return 1;
+    case first === second: return 0;
+  }
```
ivan-kadykalo, line 47: You should always have a default case

src/arrayMethodSort.js
```diff
@@ -5,8 +5,46 @@
...
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let compFunc;
+    let hadChanges;
+
+    if (arguments.length === 0) {
+      compFunc = defaultCompareFunction;
+    } else {
+      compFunc = compareFunction;
```
ivan-kadykalo, line 14: You can make it with ternary operator

`const compFunc = (arguments.length === 0) ? defaultCompareFunction : ........`

ivan-kadykalo (changes requested):
Almost done. Good job💪

src/arrayMethodSort.js
Oleksandr-Pustovit: actually I've learned another idea how to make this working, adding default function as a default value to function parameter

src/arrayMethodSort.js
```diff
@@ -4,9 +4,42 @@
...
+  const first = '' + a;
+  const second = '' + b;
+
+  switch (true) {
+    case first < second: return -1;
+    case first > second: return 1;
+    case first === second: return 0;
+    default:
+      throw new Error('some error'); // I have no idea what is it.
+  }
```
ivan-kadykalo, line 42: In the context of this function return 0 will be your default case, because you don't need to sort elements if they are the same. So you don't need that error here
```suggestion
    case first === second:
    default:
    	return 0;
  }
```

or it can work with just `default: return 0`

ivan-kadykalo (changes requested):

PR #686, opened 2022-07-12
ready

denenergi:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
```
DashaPoliazz, line 7: You can create default param function upper in code to improve code's readability

src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
...
+    String(a) > String(b) ? 1 : -1) {
+    for (let index = 1; index < this.length; index++) {
+      for (let i = 0; i < this.length - index; i++) {
+        if (compareFunction(this[i], this[i + 1]) > 0) {
+          [this[i], this[i + 1]] = [this[i + 1], this[i]];
```
DashaPoliazz, line 12: You are using `this[i]` & `this[i + 1]`. You can create variables `current = this[i]` & `next = this[i + 1]`

DashaPoliazz (approved):
Nice solution!

src/arrayMethodSort.js
ivan-kadykalo: Good point. Create a separate function, it will be good practice.

src/arrayMethodSort.js
ivan-kadykalo: Good idea to create a variable for this.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
+    String(a) > String(b) ? 1 : -1) {
+    for (let index = 1; index < this.length; index++) {
```
ivan-kadykalo, line 9: If I understand correctly, your loop has a fixed amount of iterations. By using do or while you can develop functionality when it will work till some changes are made.

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):

PR #690, opened 2022-07-13
Solution

Lentyaeev:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,25 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = standart) {
+    for (let i = 0; i < this.length - 1; i++) {
+      let temp;
+      const funcResult = compareFunction(this[i], this[i + 1]);
```
DashaPoliazz, line 10: You could create variables` const current = this[i];` & `const next = this[i + 1];` instead of repeat `this[i]` & `this[i + 1]`

DashaPoliazz (approved):
Nice solution.

PR #700, opened 2022-07-14
Solution

illnino380:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
...
+        if (compareFunction) {
+          if (compareFunction(current, previous) > 0) {
+            this[j] = current;
+            this[i] = previous;
+            current = this[i];
+          }
+          continue;
+        }
+
+        if (current.toString() > previous.toString()) {
```
VETALEK, line 23: you may move this check to the separate func and make it a default value of your arg 😉

src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      let current = this[i];
+
+      for (let j = i + 1; j < this.length; j++) {
```
VETALEK, line 11: use `do while`, so you wouldn't iterate through all the array if it is already sorted

VETALEK (changes requested):
Good start 🚀

VETALEK (approved):
🔥

PR #701, opened 2022-07-14
Added sort implementation

LovelyDrug:


chernobrovin (approved):
Good solution!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const aStr = String(a);
+    const bStr = String(b);
+
+    if (aStr > bStr) {
+      return 1;
+    } else if (aStr === bStr) {
+      return 0;
+    } else {
+      return -1;
+    }
```
VETALEK, line 17: what about `localCompare`? 😉

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+  };
+
+  [].__proto__.sort2 = function(compareFunction = compareStrings) {
+    let count;
+
+    do {
+      count = 0;
```
VETALEK, line 24: `count` is not descriptive enough

VETALEK: let's mb use some boolean like `isLastIteration`?

VETALEK (changes requested):
Good start 💪

src/arrayMethodSort.js
LovelyDrug: In this situation, localeCompare doesn't give us a result to pass our tests, because of 'ё' and 'я' characters 😢

VETALEK: image

😉

VETALEK (approved):
GJ 🔥

PR #732, opened 2022-08-11
develop

tiserett:


Kulchevych (approved):
Чудово ! 😄

src/arrayMethodSort.js
```diff
@@ -4,8 +4,29 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareString = (a, b) => {
+    return +(String(a) > String(b));
+  };
```
dellachaise, line 9: Your `compareString` function now works unproperly.  It should also return `-1` in case when `a <  b`.
Please rewrite it using `if else`.
First screen - how it should work, second - how it work now.
Screenshot 2022-08-17 at 10 58 48
Screenshot 2022-08-17 at 11 00 15

dellachaise (changes requested):
Good job 👍 

But please add changes according to comment 👇

src/arrayMethodSort.js
tiserett: fixed it

dellachaise (approved):
Well done 👍

PR #736, opened 2022-08-13
Solution

AndreyJigijalkin:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,40 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
VolodymyrChuyko, line 17:
```suggestion
  } 
    return -1;
```
This 'else' is redundant. Any way this code will run only if two previous conditions are false

src/arrayMethodSort.js
```diff
@@ -1,11 +1,40 @@
...
+      count = 0;
+
+      const prev = this[i - 1];
+      const current = this[i];
+
+      compareFunction(prev, current) > 0
+        ? (this[i - 1] = current, this[i] = prev, count++)
+        : count--;
+
+      count > 0 ? i = 0 : i;
```
VolodymyrChuyko, line 34: Would be better to use 'if' operator  instead of the ternary operator for code readability

VolodymyrChuyko (approved):
An interesting solution with a 'for' loop

src/arrayMethodSort.js
```diff
@@ -1,11 +1,40 @@
...
+  }
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
+    let count;
+
+    for (let i = 1; i < this.length; i++) {
```
AlinaMatuschak, line 24: can you use the array method instead of a simple loop?

AlinaMatuschak (approved):
🚀🚀

PR #742, opened 2022-08-14
Solution

VolodymyrChuyko:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const defaultCompare = function(a, b) {
+      switch (true) {
+        case (`${a}` > `${b}`): return 1;
+        case (`${a}` < `${b}`): return -1;
+        case (`${a}` === `${b}`): return 0;
+        default: throw new Error('Error: defaulCompre');
+      }
```
AndreyJigijalkin, line 14: I don't think that switch construction is really needed here. It looks pretty good with standard if's)

AndreyJigijalkin (approved):

src/arrayMethodSort.js
VolodymyrChuyko: It's quite good for me. So, let's agree that this is a matter of personal preferences =)

AlinaMatuschak, line 14: you can create this function outside an duse it here

src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+    const comparator = compareFunction !== undefined
+      ? compareFunction
+      : defaultCompare;
+    let isSorted;
+
+    do {
+      isSorted = true;
+
+      for (let i = 1; i < this.length; i++) {
+        const previous = this[i - 1];
```
AlinaMatuschak, line 25: can you use the array method instead of a simple loop?

AlinaMatuschak (approved):
🚀

src/arrayMethodSort.js
VolodymyrChuyko: I tried to implement .forEach method here, but that solution looks complicated and less readable to me.

PR #744, opened 2022-08-15
solution

Githi54:


KorniychukLesya (approved):
Hi! Nicely done

src/arrayMethodSort.js
```diff
@@ -1,11 +1,42 @@
...
+  }
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareString) {
+    let count;
+
+    for (let i = 1; i < this.length; i++) {
```
AlinaMatuschak, line 26: can you use the array method instead of a simple loop?

AlinaMatuschak (approved):
🚀🚀

src/arrayMethodSort.js
Githi54: I think every array methods are prohibited  in this task)

PR #745, opened 2022-08-15
Solution

Andrii-Stiebalov:


Dmytro-Matsehora (approved):
everything works as it should, good job 👍

src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
...
+    }
+
+    return this;
   };
 }
 
 module.exports = applyCustomSort;
+
+const compareFunction2 = (a, b) => {
+  if (a.toString() > b.toString()) {
```
dellachaise, line 33: this function should also return  `0` when `a.toString() === b.toString()`.

dellachaise (changes requested):
Great job 👍 

But please add changes according to comment 👇

src/arrayMethodSort.js
Andrii-Stiebalov: I'm sorry. But why? I have only one check for this output value. And for the task, I'm only interested in whether variable A is greater than variable B. If the value is 0, then it will behave the same as with a value of -1.

 I just made a review and left comments that can remove the check for _equals_ and _less_, so I have to defend my point of view. 😅

image

src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareFunction2) {
```
dellachaise, line 7: It looks strange with 2, not a good name. Better rename for example 👇 
```suggestion
  [].__proto__.sort2 = function(compareFunction = compare) {
```

src/arrayMethodSort.js
dellachaise: It is great that you could discuss and defend your opinion  👍 

Sorry, I missed 37 line of code.

dellachaise (approved):
Great job 👍

Just pay close attention to variables/functions names 😉

PR #749, opened 2022-08-17
Solution

OlehMy:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,39 @@
...
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compare) {
+    let isOut;
+
+    do {
+      isOut = false;
+
+      for (let i = 1; i < this.length; i++) {
+        if (compareFunction(this[i - 1], this[i]) > 0) {
+          [this[i - 1], this[i]] = [this[i], this[i - 1]];
```
dellachaise, line 16: Better create variables `previous` and `current` and use it.

dellachaise (approved):
Great job 👍 

Code should be easy to understand. Better create variables when it needed for this purpose 😉

PR #750, opened 2022-08-17
Solution array method sort

PastolNapas:


OleksiiPustovit (approved):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  [].__proto__.sort2 = function(callback = (a, b) =>
+    a.toString() > b.toString()) {
```
dellachaise, line 8: Better create function outside. It will be easier to read your code.

And you could also write an extended compareFunction like in [video](https://mate.academy/learn/javascript-advanced/array-iteration-methods#/video/152)

dellachaise (changes requested):
Great job 👍 

Just add a little fix 👇

src/arrayMethodSort.js
```diff
@@ -3,9 +3,29 @@
 /**
  * Implement method Sort
  */
+const callback = (a, b) => a.toString() > b.toString();
```
dellachaise, line 6: In this function you could also return `-1`.

dellachaise (approved):
Great job 🔥

PR #752, opened 2022-08-17
solution

anyakrainiuk:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,40 @@
...
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
+
```
Andrii-Stiebalov, line 18: Why do you need as many as three different output options, if you only care about more than 0 or not? 🤓

This will be enough 😉👇
 ```
 if (stringA > stringB) {
    return 1;
```

image

**Review by Andrii-Stiebalov — APPROVED** [review]
Good job!!! 🚀👏

File: src/arrayMethodSort.js [hunk 2]
```diff
@@ -3,10 +3,36 @@
...
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsString) {
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
```
AlinaMatuschak, line 22: can you use the array method instead of a simple loop?

AlinaMatuschak (approved):
🔥

PR #764, opened 2022-08-22
Solution

MihailKostov:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const sortMethod = (a, b) => {
```
SanyaBratashchuk, line 7:
```suggestion
  const defaultCompareFunction = (a, b) => {
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    return (stringA > stringB) ? 1
+      : (stringA === stringB) ? 0
+        : -1;
```
SanyaBratashchuk, line 13: Don't use nested ternary operators. It is hard to read

SanyaBratashchuk (changes requested):

MihailKostov:
So is it better to use ternary operators or operator If?

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
valerii-derkach, line 17:
```suggestion
    if (stringA > stringB) {
      return 1;
    } 
    if (stringA < stringB) {
      return -1;
    } 
    return 0;
```

valerii-derkach (approved):

PR #765, opened 2022-08-23
solved

PaulVoron:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
```
Bohdan-Maksymiuk, line 40: avoid such nested `else if` conditions
This code is hard to read and maintain

Bohdan-Maksymiuk (approved):

PR #779, opened 2022-09-09
Solution

illia-kots:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
SanyaBratashchuk, line 17: you don' t need `else if` and `else` if you have `return` statement in your `if`

SanyaBratashchuk (changes requested):

src/arrayMethodSort.js
illia-kots: Ok

SanyaBratashchuk (approved):

PR #783, opened 2022-09-13
Solution

MaxManis:


Denchepornyuk (changes requested):
your method works differently than the original method

Screenshot 2022-09-14 at 11 33 30

Denchepornyuk (approved):

PR #787, opened 2022-09-17
sort method solution

TheVovchik:


iPanda177 (approved):
Well done!)

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
imaksweb, line 40:
```suggestion
  if (stringA > stringB) {
    return 1;
  }
  if (stringA === stringB) {
    return 0;
  }
  return -1;
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
 
+function compareAsStrings(a, b) {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  }
```
Anna-Tkachenko, line 38: else is redundant after return statement

Anna-Tkachenko (approved):

src/arrayMethodSort.js
TheVovchik: will split it

PR #791, opened 2022-09-19
sort implemented

varkovroma:


grafit-off (approved):
Good job!

src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
silchyn, line 17:
```suggestion
  return stringA.localeCompare(stringB);
```

silchyn (changes requested):

src/arrayMethodSort.js
silchyn: ?

silchyn (changes requested):
Do you have some counterarguments or what?

PR #792, opened 2022-09-19
js_array-method-sort

xiimiik:


Yevheniya-Deryka (approved):
GJ! 
But as to me it were easier to understand what is happening here when default function is brought to separate function

src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compare = (prev, curr) => (String(prev) > String(curr))) {
```
OleksiiBlezniuk, line 8:
```suggestion
  [].__proto__.sort2 = function(
    compare = (prev, curr) => (String(prev) > String(curr))
  ) {
```

OleksiiBlezniuk: At least wrap closing bracket, but better to move it to separated constant

OleksiiBlezniuk (approved):

PR #794, opened 2022-09-19
Custom sort method was implemented

Yevheniya-Deryka:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
xiimiik, line 39:
```suggestion
  return stringA.localeCompare(stringB);
```

xiimiik (changes requested):
Nice work! Left one comment
I think it will be shorter

src/arrayMethodSort.js
Yevheniya-Deryka: it won't work in my case because the logic is build on the compareFunction should return a value greater than zero, but not true or false value

OleksiiBlezniuk: @xiimiik Also default sort method compares strings by charcode, not like localeCompare
Screenshot 2022-09-20 at 17 57 29

OleksiiBlezniuk, line 39: You've used return here, remove this `else`s

OleksiiBlezniuk (changes requested):

src/arrayMethodSort.js
Yevheniya-Deryka: done

PR #801, opened 2022-09-20
solutin

miamarichka:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
SanyaBratashchuk, line 16: you don't need else if/else when you have return statement inside your if

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #804, opened 2022-09-20
Method Sort

Anna-Borodaienko:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+          [this[i], this[i + 1]] = [this[i + 1], this[i]];
+          count++;
+        }
+      }
+    }
+
+    return this;
   };
+
+  function compareFuncAsOriginalArraySortMethod(firstElement, secondElement) {
```
jojnajbude, line 27: So long name for function

jojnajbude (approved):

src/arrayMethodSort.js
OleksiiBlezniuk, line 27: epic name 🙂

OleksiiBlezniuk: but better to use something shorter

OleksiiBlezniuk: for example `initialCompareFn` or smth like that

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  compareFuncAsOriginalArraySortMethod) {
```
OleksiiBlezniuk, line 8:
```suggestion
  [].__proto__.sort2 = function(
    compareFunction = compareFuncAsOriginalArraySortMethod
  ) {
```

OleksiiBlezniuk: Open and closing brackets should be on same column, inner content should have 2 spaces indentation from id (just like in html)

OleksiiBlezniuk (approved):
Create meaningful but also laconic names 🙂

PR #805, opened 2022-09-21
Solution

Oleksandra-Shevchenko:


DenysKleimenov (approved):
Good!

src/arrayMethodSort.js
```diff
@@ -3,10 +3,42 @@
...
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
```
naziamoff, line 42: you don't need `if...else` construction here, you can avoid it where possible.
We had this on the checklist in some of our previous tasks

naziamoff (changes requested):
LGTM 🔥 
Just a small comment to fix

naziamoff (approved):
🔥

PR #806, opened 2022-09-21
Custom implementation of Array.prototype.sort()

Vladosik1337:


illia-sosnytskyi (commented):
GJ, nice solution)

src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = ((a, b) => (
+    a.toString() > b.toString()
+  ))) {
```
Anna-Tkachenko, line 9: will be better to write this function using function declaration separately from function arguments
also, you need to write default sort functionality (function should return 0 if elements are the same) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Anna-Tkachenko (changes requested):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,36 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (compareFunction(this[i], this[j]) > 0) {
+          const swap = this[i];
+
+          this[i] = this[j];
+          this[j] = swap;
+        }
+      }
+    }
```
Anna-Tkachenko, line 33: let's rewrite it using do while cycle
method sort should go through the elements one more time only if on previous array iteration there was 1 or more elements changing

Anna-Tkachenko (changes requested):

denys-cheporniuk (approved):

PR #811, opened 2022-09-21
Solution

andrew-satyr:


varkovroma (approved):
GJ

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    }
+
+    return -1;
+  }
```
naziamoff, line 18: no need in `if...else` here. You can do it with help of 2 `if's`

naziamoff (changes requested):
GJ, but just one small thing

naziamoff (approved):
🔥

PR #814, opened 2022-09-21
solution

artem-musii:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,41 @@
...
+  function defaultComparison(a, b) {
+    if (a.toString() < b.toString()) {
+      return -1;
+    }
+
+    if (a.toString() === b.toString()) {
+      return 0;
+    }
+
+    if (a.toString() > b.toString()) {
```
anastasiia-nahaiska, line 38: I think it condition isn't necessary

src/arrayMethodSort.js
```diff
@@ -4,9 +4,41 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultComparison) {
+    let cnt = 1;
```
anastasiia-nahaiska, line 8: As for me `count` is more understandable, but maybe `cnt` is often practiced)

anastasiia-nahaiska (approved):
GJ!

src/arrayMethodSort.js
Anna-Tkachenko: agree, fix variable name, cnt - bad name

src/arrayMethodSort.js
Anna-Tkachenko: yes, you can return one of values by default

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #815, opened 2022-09-21
custom sort method was created

Andrew-Skoryk:
Solution

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -4,9 +4,39 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareString) {
+    let count = 1;
+
+    while (count > 0) {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const currentValue = this[i];
+        const prevValue = this[i - 1];
+
+        if (compareFunction(prevValue, currentValue) > 0) {
+          this[i - 1] = currentValue;
+          this[i] = prevValue;
+          count++;
+        }
+      }
+    }
+
+    return this;
   };
 }
+
+const compareString = (a, b) => {
+  const strA = `${a}`;
+  const strB = `${b}`;
+
+  if (strA > strB) {
+    return 1;
+  } else if (strA < strB) {
+    return -1;
+  } else {
+    return 0;
+  }
+};
 
 module.exports = applyCustomSort;


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+      }
+    }
+
+    return this;
   };
 }
 
+const compareString = (a, b) => {
+  const strA = `${a}`;
+  const strB = `${b}`;
```
dmytristepanov, line 31: You can name arguments more understandable

dmytristepanov (approved):
Excellent work!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const strPrev = `${prevValue}`;
+  const strCurrent = `${currentValue}`;
+
+  if (strPrev > strCurrent) {
+    return 1;
+  } else if (strPrev < strCurrent) {
+    return -1;
+  } else {
+    return 0;
+  }
```
naziamoff, line 39: You should avoid else if it is not necessary
```suggestion
  if (strPrev > strCurrent) {
    return 1;
  }  
  
  if (strPrev < strCurrent) {
    return -1;
  } 
    
  return 0;
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+      }
+    }
+
+    return this;
   };
 }
 
+const compareString = (prevValue, currentValue) => {
+  const strPrev = `${prevValue}`;
+  const strCurrent = `${currentValue}`;
```
naziamoff, line 31: Think of different way of type conversion, using interpolation is definitely not the best decision here

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareString) {
+    let count = 1;
```
naziamoff, line 8: if you use `do...while` loop instead of while - you won't need to set 1 as initial for `count`.
Because 1 looks like some magic number here, we should avoid this

naziamoff (changes requested):

naziamoff (changes requested):
You did not fix `if...else` issue

src/arrayMethodSort.js
```diff
@@ -4,9 +4,41 @@
...
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+const compareString = (prevValue, currentValue) => {
+  const strPrev = prevValue + '';
+  const strCurrent = currentValue + '';
```
naziamoff, line 31: This one is better
```suggestion
  const strPrev = String(prevValue);
  const strCurrent = String(currentValue);
```

naziamoff (approved):
Well done, but check my comment about converting value to string

PR #827, opened 2022-09-23
solution

Naveed-Shoukat:


kamal-tufekcic (approved):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
...
+          : compareFunction(this[i], this[j]) > 0;
+
+        if (swapCheck) {
+          [
+            this[i],
+            this[j],
+          ] = [
+            this[j],
+            this[i],
+          ];
```
VinnieJ-2k20, line 24: Avoid using destructuring like this, it's extremely hard to read.

src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      let currentElement = String(this[i]);
+
+      for (let j = i + 1; j < this.length; j++) {
+        const nextElement = String(this[j]);
+        const swapCheck = compareFunction === undefined
```
VinnieJ-2k20, line 13: Can introduce default compare function.

VinnieJ-2k20 (changes requested):
Some inprovements can be made.

src/arrayMethodSort.js
Naveed-Shoukat: compare function introduced

src/arrayMethodSort.js
```diff
@@ -5,7 +5,28 @@
...
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const defaultCompare = (firstElement, secondElement, compareFn) => {
+      if (compareFunction) {
+        return compareFn(firstElement, secondElement) > 0;
+      }
+
+      return String(firstElement) > String(secondElement);
+    };
```
VinnieJ-2k20, line 14: Would be better to declare it outside and use as default value for compareFunction.

VinnieJ-2k20 (approved):

PR #829, opened 2022-09-24
Solution

rstarovoit:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
 /**
  * Implement method Sort
  */
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
```
grafit-off, line 8: Why A and B? You can use prev and current, it's more meaningful

grafit-off (approved):
GJ 🔥

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
```
naziamoff, line 17: you can avoid `else` in this block, try to manage it only with `if's`

src/arrayMethodSort.js
naziamoff: I would think of something like `prevItem` and `currentItem` here

naziamoff (changes requested):

PR #831, opened 2022-09-28
first commit

viktoria-krasnyuk:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringB = String(b);
+
+  if (srtingA > stringB) {
+    return 1;
+  } else if (srtingA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
```
SanyaBratashchuk, line 17: if you have `return` statement inside it isn't neccessary to use `else if`

SanyaBratashchuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #838, opened 2022-09-29
solution

oleksii-bieliaiev:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const strA = String(a);
+    const strB = String(b);
+
+    if (strA > strB) {
+      return 1;
+    } else if (strA === strB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
SanyaBratashchuk, line 17: you don't need `else if` and `else` if you have `return` inside `if` block

SanyaBratashchuk (approved):

PR #840, opened 2022-09-30
Solution

Fispil:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
SanyaBratashchuk, line 16: you don't need `else if` or `else` if you have `return` in `if` block

SanyaBratashchuk (changes requested):

liliia-livandovskaya (approved):
Good job! 🔥

PR #842, opened 2022-10-04
done

studyHITHUB:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,38 @@
...
+  const strA = String(a);
+  const strB = String(b);
+
+  if (strA > strB) {
+    return 1;
+  } else if (strA === strB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
AlisaVlasova, line 13: you don't need else statements if you return value there

AlisaVlasova (changes requested):

src/arrayMethodSort.js
```diff
@@ -1,11 +1,34 @@
...
 
-/**
- * Implement method Sort
- */
+const compareFunction = (a, b) => {
+  const strA = String(a);
+  const strB = String(b);
+
+  if (strA > strB) {
+    return 1;
```
TarasHoliuk, line 8: I guess you misunderstood this [comment](https://github.com/mate-academy/js_array-method-sort/pull/842#discussion_r990350108)

You still need cases that return `0` or `-1`
But better to format it in another way (without `else` keyword)
just return `-1` at the end of the function outside `if` statement

TarasHoliuk (changes requested):

TarasHoliuk (approved):

PR #845, opened 2022-10-07
function applyCustomSort was added

Olena-Shevchenko:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
liliia-livandovskaya, line 39:
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA === stringB) {
    return 0;
  }
  
  return -1;
  
  }
```

liliia-livandovskaya (changes requested):
GJ! Lets work on fix

SanyaBratashchuk (approved):

PR #856, opened 2022-10-20
sorted without eslint no-param

metinbicaksiz:
disbaled the no-param reassign property :)

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -3,9 +3,43 @@
 /**
  * Implement method Sort
  */
+function compareAsString(a, b) {
+  /* eslint-disable no-param-reassign */
+  a = String(a);
+  b = String(b);
+
+  if (a > b) {
+    return 1;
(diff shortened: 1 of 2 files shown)


src/arrayMethodSort.js
```diff
@@ -3,9 +3,43 @@
 /**
  * Implement method Sort
  */
+function compareAsString(a, b) {
+  /* eslint-disable no-param-reassign */
```
amitpatiljc, line 7: nop you should not do that
You can use other variable and reassign

src/arrayMethodSort.js
```diff
@@ -3,9 +3,43 @@
...
+  a = String(a);
+  b = String(b);
```
amitpatiljc, line 9:
```suggestion
 const element1 = String(a);
 const element2 = String(b);
```

amitpatiljc (changes requested):

PR #864, opened 2022-10-29
Solution

Illia-Skladnik:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,38 @@
...
- */
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
```
iryna-lelyk, line 11: When you have if condition with return, better to write if from new line (not if...else if... else, but if, if, if). This rule was in checklist to task 'transportation on vacation'. 
https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md

iryna-lelyk (approved):

src/arrayMethodSort.js
Illia-Skladnik: Yes, thanks!

PR #865, opened 2022-10-29
solution

IOneStarI:


taraselnik (approved):
cant add anything enjoy

src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+function compareStrings(previos, next) {
+  const previosString = String(previos);
+  const nextString = String(next);
+
+  if (previosString > nextString) {
+    return 1;
+  }
+
+  return -1;
+}
```
denys-danyliuk, line 39: move it to the top of the file, and make it arrow function

src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+function compareStrings(previos, next) {
+  const previosString = String(previos);
```
denys-danyliuk, line 31:
```suggestion
  const previousString = String(previous);
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+  const nextString = String(next);
+
+  if (previosString > nextString) {
+    return 1;
+  }
+
+  return -1;
```
denys-danyliuk, line 38: but what to do if they equal?🤔

denys-danyliuk (changes requested):

PR #866, opened 2022-10-29
Solution

IhorChaikovskyi:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,39 @@
...
+  const firstStrElement = String(firstElment);
+  const secondStrElement = String(seondElement);
+
+  if (firstStrElement > secondStrElement) {
+    return 1;
+  } else if (firstStrElement < secondStrElement) {
+    return -1;
+  } else {
+    return 0;
+  }
```
TarasHoliuk, line 13: taking into account that you are doing `return` in each condition better rewrite it without `else if` and `else`
the result will be more readable 
generally: try to avoid `else if` and `else` if it's possible

TarasHoliuk (changes requested):
Good job! I've added one suggestion

src/arrayMethodSort.js
IhorChaikovskyi: Ok. Thanks

src/arrayMethodSort.js
```diff
@@ -1,11 +1,40 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
+function toStringSort(firstElment, seondElement) {
+  const firstStrElement = String(firstElment);
+  const secondStrElement = String(seondElement);
+
+  switch (true) {
```
TarasHoliuk, line 7: It's the wrong usage of `switch` construction
Again, better to use simple `if` statements here :)

TarasHoliuk (changes requested):

src/arrayMethodSort.js
IhorChaikovskyi: ok, i got it

Virass (approved):

PR #868, opened 2022-10-29
develop

MaxFedak:


masha-girya (approved):
good job!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,42 @@
...
+    if (prevEl > curEl) {
+      return 1;
+    }
+
+    if (prevEl < curEl) {
+      return -1;
+    }
+
+    return 0;
   };
```
denys-danyliuk, line 43: create arrow functions before you use them

denys-danyliuk (changes requested):

src/arrayMethodSort.js
MaxFedak: but functions are instantly fully initialized, or it's a rule of good code?

denys-danyliuk: it's rule of good code and for better readability to create all functions at the top

PR #875, opened 2022-10-31
solution

taraselnik:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,32 @@
...
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let internalExternalSwitchFunc = compareFunction;
+
+    if (internalExternalSwitchFunc === undefined) {
+      internalExternalSwitchFunc = (a, b) => {
+        if (`${a}` > `${b}`) {
+          return 1;
+        } else if (`${a}` === `${b}`) {
+          return 0;
```
andrii-bassarab, line 15: ```if (`${a}` === `${b}`) {
return 0;
}
```

Comment (andrii-bassarab) [inline src/arrayMethodSort.js:15]:
```suggestion
        if (`${a}` === `${b}`) {
          return 0;
```

andrii-bassarab (approved):
Good job, interesting solution!

src/arrayMethodSort.js
taraselnik: thanks, a lot!

denys-danyliuk (approved):

PR #877, opened 2022-11-01
Solution

andriy-shymkiv:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,38 @@
...
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
+
+[].__proto__.sort2 = function(callback = applyCustomSort) {
+  let count;
```
karyna-salnykova, line 20: Why won't you write count = 0 at this moment? Then there's no need to write it in do-while

referee1988 (approved):
Great job!🔥

src/arrayMethodSort.js
```diff
@@ -3,10 +3,38 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
denys-danyliuk, line 16:
```suggestion
  } 
    return -1;
  }
```
else is redundant here

denys-danyliuk (approved):

PR #882, opened 2022-11-02
Solution

alpopkov:


Stepico (approved):
It's good, but I guess this solution with 2 nested fors without any counter that can interrupt this procedure is worse than if you have one. Just imagine how many redundant operations your code should do if you have an array of million elements and they've already  been ordered

src/arrayMethodSort.js
```diff
@@ -4,8 +4,32 @@
...
+    } else if (aString === bString) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
+
+  [].__proto__.sort2 = function(compareFunction = myCompareFunction) {
+    for (let i = 0; i < this.length; i++) {
+      for (let y = 1; y < this.length; y++) {
```
denys-danyliuk, line 22: this is wrong implementation of how method sort should work, read more [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)

denys-danyliuk (changes requested):

PR #885, opened 2022-11-03
complete

spiritkyiv:


Largoxoxo (approved):
Nested cycles are EVIL! JK^^
GJ, very nice solution!

src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let x = 0; x < this.length; x++) {
+      for (let y = 0; y < (this.length - x - 1); y++) {
```
denys-danyliuk, line 9: try to find better solution

src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
...
+        const currentIndex = String(this[y]);
+        const nextIndex = String(this[y + 1]);
+
+        let compareResult = false;
+
+        if (typeof (compareFunction) === 'function') {
```
denys-danyliuk, line 15: why do you need this check?

denys-danyliuk (changes requested):
Read more about how it should work [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)
You need to implement 'compareFunction' as well

Olexandr6 (approved):

PR #892, opened 2022-11-06
Solution

ndv20:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+function compareAsString(a, b) {
+  const strA = String(a);
+  const strB = String(b);
+
+  if (strA > strB) {
+    return 1;
+  } else if (strA === strB) {
+    return 0;
+  } else {
+    return -1;
```
polosanya, line 37: Avoid nesting if...else.
Try to use switch case instead
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsString) {
+    let count, temp;
```
polosanya, line 8: Try to use more informative name than `temp`

polosanya (changes requested):
Good job! 
Check my comments to improve your code

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const strB = String(b);
+
+  switch (true) {
+    case (strA > strB):
+      return 1;
+    case (strA === strB):
+      return 0;
+    default:
+      return -1;
+  }
```
polosanya, line 39: I'm sorry I confused you with my previous comment, but nevertheless, it's better to use `if` statements here. Just no need to use `else`, as your code will not go further after `return` anyway. What about switch...case - it's a great tool when you are switching some variable, but it's better to avoid `switch(true)` as it works the same as if...else  but can be more difficult for other developers to understand.
```suggestion
   if (strA > strB) {
      return 1;
   }
   
   if (strA < strB) {
     return -1;
   }
   
   return 0;
```

polosanya (approved):
Good job! Please check my comment

PR #893, opened 2022-11-07
Solution

SerhiiVovchenko:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
 /**
  * Implement method Sort
  */
+const compareAsStr = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
```
polosanya, line 12: Try to avoid nesting if...else
Use switch case instead

polosanya (changes requested):
Good job! Check my comment

liliia-livandovskaya (approved):
Great job 🔥

PR #895, opened 2022-11-08
Method sort

maksym-kryvolap:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
```
tetiana-ivanchenko, line 18:
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  if (stringA === stringB) {
    return 0;
  }
  
  return -1;
};
```

src/arrayMethodSort.test.js
```diff
@@ -55,6 +55,6 @@ test('custom compare function, numbers: [3, 12, 2, 11].sort2((a, b) => a - b)',
 test(
   `custom compare function, strings: 
   ['ёж','як','аист'].sort2((a, b) => a.localeCompare(b))`, () => {
-    expect(['ёж', 'як', 'аист'].sort2((a, b) => a.localeCompare(b)))
+    expect(['ёж', 'як', 'аист'].sort2((a, b) => a. localeCompare(b)))
```
tetiana-ivanchenko, line 58: revert this change

src/arrayMethodSort.js
tetiana-ivanchenko: When you have if condition with return, no need to write if else.

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #900, opened 2022-11-10
solution

euFarafonov:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,41 @@
...
+      if (strA > strB) {
+        return 1;
+      }
+
+      if (strA < strB) {
+        return -1;
+      }
+
+      return 0;
+    }
```
liliia-livandovskaya, line 21:
```suggestion
    function compareStr(a, b) {
      const strA = String(a);
      const strB = String(b);

      if (stringA === stringB) {
    return 0;
  }

  return stringA > stringB ? 1 : -1;
  
    }
```
let's move the `compareStr` function out from the `applyCustomSort`

liliia-livandovskaya (changes requested):
GJ! Pay attention to the comments

liliia-livandovskaya (approved):
Awesome 🔥

PR #901, opened 2022-11-10
array method sort created

katya-suvorova:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,28 @@
...
  */
+
+function compareString(a, b) {
+  if (String(a) > String(b)) {
+    return 1;
+  } else if (String(a) < String(b)) {
+    return -1;
+  } else {
+    return 0;
+  }
```
liliia-livandovskaya, line 14: const stringA = String(a);
const stringB = String(b);

return (stringA > stringB) ? 1 : (stringA === stringB) ? 0 : -1;

src/arrayMethodSort.js
```diff
@@ -3,9 +3,28 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareString) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (compareFunction(this[i], this[j]) > 0) {
+          [this[i], this[j]] = [this[j], this[i]];
+        }
+      }
+    }
```
liliia-livandovskaya, line 25: let's use `do while` instead of nested loops

liliia-livandovskaya (changes requested):
Good job! Pay attention to the comments

src/arrayMethodSort.js
katya-suvorova: updated

src/arrayMethodSort.js
katya-suvorova: updated

polosanya (approved):
Well done!

PR #905, opened 2022-11-15
solution

olenitut:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let toRepeat = true;
+
+    const bubbleSort = () => {
+      const defaultCheck = (a, b) => {
```
Viktor-Kost, line 11: You can move to separated variable this `defaultCheck` function (service function) to make your code clear

src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
...
+        if (String(a) > String(b)) {
```
Viktor-Kost, line 12: Here is a good situation to use ternary operator [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)

src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
...
+          return 1;
+        }
+
+        return -1;
+      };
+
+      const checkCondition = compareFunction || defaultCheck;
```
Viktor-Kost, line 19: We can avoid this line of code by adding  `defaultCheck` as a default parameter to `sort2 `method
Like this `function(compareFunction = defaultCheck)`

Viktor-Kost (changes requested):

Viktor-Kost (approved):
Nice 👍

PR #912, opened 2022-11-21
Solution

KrisMakarovska:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
liliia-livandovskaya, line 17:
```suggestion
    if (stringA > stringB) {
      return 1;
    } 
    
    if (stringA === stringB) {
      return 0;
    } 
    
    return -1;
    
```

liliia-livandovskaya: that is not a mistake but it would be great to make the code more readable

liliia-livandovskaya (approved):
Great job! 🔥

PR #913, opened 2022-11-21
Solution

Yevhenii-Stanchenko:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,52 @@
...
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
 module.exports = applyCustomSort;
+// function sortFunction(a, b){
```
TarasHoliuk, line 42: don't leave comments

src/arrayMethodSort.js
```diff
@@ -3,10 +3,52 @@
...
  */
+function compareStrings(previos, next) {
+  const previosElem = String(previos);
+  const nextElem = String(next);
+
+  if (previosElem > nextElem) {
+    return 1;
+  }
+
+  return -1;
```
TarasHoliuk, line 14: it's better to return `0` if values are equal (explicitly)

nazarmatsevych (approved):

PR #916, opened 2022-11-24
Solution

UmizDemud:
It seemed to me that there is a bug in the tests, I already reported it.

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -5,7 +5,43 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let count;
+    let current;
+    let prev;
+
+    if (!compareFunction) {
+      do {
+        count = 0;
+
+        for (let i = 1; i < this.length; i++) {
+          current = this[i];
+          prev = this[i - 1];
+
+          if (current.toString().localeCompare(prev.toString()) < 0) {
+            this[i] = prev;
+            this[i - 1] = current;
+            count++;
+          }
+        }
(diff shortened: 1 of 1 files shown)


src/arrayMethodSort.js
```diff
@@ -5,7 +5,43 @@
...
+        for (let i = 1; i < this.length; i++) {
+          current = this[i];
+          prev = this[i - 1];
+
+          if (compareFunction(current, prev) < 0) {
+            this[i] = prev;
+            this[i - 1] = current;
+            count++;
+          }
+        }
```
mgrinko, line 40: this code is repeated twice
please use the default value for the `compareFunction` to remove the duplication

mgrinko (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,43 @@
...
+
+      if (currentDif > 0) {
+        return 1;
+      } else if (currentDif === 0) {
+        continue;
+      } else {
+        return -1;
+      }
+    }
+  };
```
meforoma, line 23: i think this can be way easier - simply compare strings
```
if (String(a) > String(b)) { return ?
if (String(a) < String(b)) { return ?
return ?
```

meforoma (changes requested):
default compare is way easier 🙂 try that)

src/arrayMethodSort.js
UmizDemud: I forgot about the default comparisons of these level libraries.  Correcting

naziamoff (approved):
👍

PR #919, opened 2022-11-26
Solution

vadiimvooo:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+          count++;
+        }
+      }
+    } while (count);
+
+    return this;
   };
 }
 
+const compareString = function compareItems(a, b) {
```
ivan-kadykalo, line 29: You can just create your function, why do you assign it to the new variable?

```suggestion
function compareString(a, b) {
```

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #923, opened 2022-11-28
solution

lefteris-L:


nurdanokcu (approved):
Well done. :rocket:

src/arrayMethodSort.js
```diff
@@ -5,8 +5,23 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const haveArgs = arguments.length > 0;
+
+    for (let iterator = this.length - 1; iterator >= 0; iterator--) {
+      for (let i = 0; i < iterator; i++) {
```
meforoma, line 11: better avoid nested `for` cycles

src/arrayMethodSort.js
```diff
@@ -5,8 +5,23 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
meforoma, line 7: it would be a really good idea to introduce a default sorting function
and use it as a default variable
```suggestion
  [].__proto__.sort2 = function(compareFunction = defaultCompareFunction) {
```

src/arrayMethodSort.js
```diff
@@ -5,8 +5,23 @@
...
-    // write code here
+    const haveArgs = arguments.length > 0;
+
+    for (let iterator = this.length - 1; iterator >= 0; iterator--) {
+      for (let i = 0; i < iterator; i++) {
+        const current = this[i];
+        const next = this[i + 1];
+
+        if ((haveArgs && compareFunction(current, next) > 0)
+        || (!haveArgs && `${current}` > `${next}`)) {
```
meforoma, line 16: it's not really readable and not easy to understand what's done in this line
that's where a `defaultCompareFunction` would be very useful in particular

src/arrayMethodSort.js
```diff
@@ -4,9 +4,30 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => `${a}` > `${b}`) {
```
meforoma, line 7: this is wrong )
the `.sort` callback expected output is a number `-1 | 0 | 1`, right?
this would not return the expected outputs
```
`${a}` > `${b}`
```
introduce a proper separate helper function:
```
function defaultSort(...) {...}// returns `-1 | 0 | 1`

// then use it as default
function(compareFunction = defaultSort) {...}
```

even if you'd need to compare strings, why not just use `single quotes` or even better - a `String` class:
`'a' > 'b'` || `String(a) - String(b)`

lefteris-L: single quotes comparison doesn't seem to work. Throws an eslint line `Unexpected constant condition.eslint[no-constant-condition]`

PR #924, opened 2022-11-29
Solution

nurdanokcu:


UmizDemud (approved):
Good job 👍

meforoma (approved):
👍 default compare function
👍 do...while
👍 swap elements in-place

PR #925, opened 2022-11-29
arrayMethodSort

ysrckr:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,31 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultSort) {
+    let temp;
+
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
```
UmizDemud, line 11:
```js
swapped = false;

File: src/arrayMethodSort.js [hunk 2]
```diff
@@ -4,9 +4,31 @@
...
+        if (compareFunction(this[i], this[j]) > 0) {
```
UmizDemud, line 12:
```js
swapped = true;

File: src/arrayMethodSort.js [hunk 3]
```diff
@@ -4,9 +4,31 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultSort) {
+    let temp;
```
UmizDemud, line 8:
```js
let swapped = true;

File: src/arrayMethodSort.js [hunk 4]
```diff
@@ -4,9 +4,31 @@
...
+
+    for (let i = 0; i < this.length; i++) {
```
UmizDemud, line 10: `while (swapped)` instead

UmizDemud (changes requested):
If you write this with the same inner 'for' but with while outside (with a control variable), it'd be efficient

UmizDemud (approved):
Makes sense

You can implement the while(swapped) issue for more efficiency 👍

src/arrayMethodSort.js
```diff
@@ -4,9 +4,31 @@
...
 }
 
+function defaultSort(a, b) {
+  if (String(a) > String(b)) {
+    return 1;
+  } else if (String(a) < String(b)) {
+    return -1;
+  } else {
+    return 0;
+  }
```
meforoma, line 31: no need for `else` nesting when `if` has a `return` statement
```suggestion
  if (String(a) > String(b)) {
    return 1;
  }
  
  if (String(a) < String(b)) {
    return -1;
  }
  
  return 0;
```

src/arrayMethodSort.js
meforoma, line 11: you don't need the nested `if` - just use `i + 1` instead of `j`

meforoma (changes requested):
please see the comments

PR #927, opened 2022-11-29
sort solution

mr-valentos:


GabrielMaksimovich (approved):
everyrhing like Misha made, nothing to add

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+
+    if (aToString > bToString) {
+      return 1;
+    } else if (aToString === bToString) {
+      return 0;
+    } else if (aToString < bToString) {
+      return -1;
+    }
+  };
+
```
valerii-derkach, line 19: if you return value, no need else here, and also no need to compare  aToString === bToString , just use return 0 in the end

valerii-derkach (changes requested):

solaryasha (approved):
👍

ivan-kadykalo (approved):

PR #935, opened 2022-12-03
solution

KrisPlanka:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
   };
```
ivan-kadykalo, line 40: Let's move the helping function outside the main function.

PR #942, opened 2022-12-04
Solution

partnersinbahamas:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const strA = String(a);
+    const strB = String(b);
+
+    if (strA > strB) {
+      return 1;
+    } else if (strA === strB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
tetiana-ivanchenko, line 17:
```suggestion
    if (strA > strB) {
      return 1;
    }
    
    if (strA === strB) {
      return 0;
    }
    
    return -1;
```

tetiana-ivanchenko: if you have return you don't need to use else

PR #945, opened 2022-12-05
Solution

batstolya:


kreisberg (approved):
Seems great)

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA < stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
```
valerii-derkach, line 17:
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  if (stringA < stringB) {
    return 0;
  }
  
   return -1;
};
```

no need` else` statement when we `return` value inside` if `block

valerii-derkach (approved):

PR #949, opened 2022-12-05
finished

jackelashkar:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,36 @@
...
+  [].__proto__.sort2 = function(compareFunction = compareDefault) {
+    for (let i = 1; i < this.length; i++) {
+      for (let j = i; j > 0; j--) {
+        const prev = this[j - 1];
+        const current = this[j];
+
+        if (compareFunction(prev, current) > 0) {
+          this[j - 1] = current;
+          this[j] = prev;
+        } else {
```
UmizDemud, line 29: You don't need to break here, it is the end of the scope already. Just kinda makes the code longer

UmizDemud (changes requested):
Seems perfect, apart from the comment. Just let me know when you do the correction, I can approve right away.

src/arrayMethodSort.js
jackelashkar: thank you Umut for clarifying that out :)

src/arrayMethodSort.js
```diff
@@ -3,9 +3,34 @@
...
+    return -1;
+  }
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareDefault) {
+    for (let i = 1; i < this.length; i++) {
+      for (let j = i; j > 0; j--) {
```
naziamoff, line 22: nested for loops is 🙅

naziamoff (changes requested):

naziamoff (approved):

PR #951, opened 2022-12-06
Solution

Anastasia4sik:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+const defaultCompare = (a, b) => {
+  const aStr = String(a);
+  const bStr = String(b);
```
llllewvvaa, line 31: I think, no need to create this vars here
```suggestion
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+
+  if (aStr > bStr) {
+    return 1;
+  } else if (aStr === bStr) {
+    return 0;
+  }
+
+  return -1;
```
llllewvvaa, line 39: You can compare string actually in the `if` statement.
And avoid `else if`, by adding another if and changing `return` value
```suggestion
  if (String(a) > String(b)) {
    return 1;
  }
  
  if (String(a) < String(b)) {
    return -1;
  }
  
  return 0;
```

llllewvvaa (changes requested):
It's your choice to create `variables` where I marked or not, but I see them like that, just correct if statement :)

src/arrayMethodSort.js
Anastasia4sik: I'd better keep these variables to avoid repeating myself and changing the parameters to String type one more time

llllewvvaa (approved):

src/arrayMethodSort.js
```diff
@@ -4,9 +4,41 @@
...
+          count++;
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+const defaultCompare = (a, b) => {
```
ivan-kadykalo, line 29: Arrow functions should be declared before usage. Move this function before `applyCustomSort()`

ivan-kadykalo (approved):
Good job 🔥

PR #957, opened 2022-12-07
solution

vladyslav-krainyk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareAsStrings = (a, b) => {
+    switch (true) {
```
stanislav-topikha, line 8: `switch true` is a bad practice

stanislav-topikha (changes requested):

stanislav-topikha (approved):

PR #959, opened 2022-12-09
Solution

skomskiy:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
Viktor-Kost, line 17: It's better not to use nesting else if condition
Use an early return as a variant

PR #964, opened 2022-12-12
Solution

IrynaK-a:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA < stringB) {
+      return -1;
+    } else {
+      return 0;
+    }
```
polosanya, line 17:
```suggestion
    if (stringA > stringB) {
      return 1;
    }
    
    if (stringA < stringB) {
      return -1;
    }
    
    return 0;
```
There is no need to use else...if with return. Your code will not go further after return anyway
https://medium.com/@rachelchervin/if-else-vs-if-return-using-implicit-returns-in-javascript-and-react-bbac63656e1b

polosanya (changes requested):
Almost done! Please check the comment

vlad-demchuk (approved):

PR #965, opened 2022-12-12
first commit

haduigon:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,42 @@
...
+    const bString = String(b);
+
+    if (aString > bString) {
+      return 1;
+    } else if (aString === bString) {
+      return 0;
+    } else {
+      return -1;
+    }
+  }
```
tetiana-ivanchenko, line 18:
```suggestion
    if (aString > bString) {
      return 1;
    }
     if (aString === bString) {
      return 0;
    }
    
    return -1;
  }
```

tetiana-ivanchenko: if you return smth in the if block - else after it is redundant

src/arrayMethodSort.js
```diff
@@ -4,8 +4,42 @@
...
+
+  [].__proto__.sort2 = function(compareFunction = compareDefault) {
     // write code here
+
+    let zero = this[0];
+    let first = this[1];
+    let count;
```
tetiana-ivanchenko, line 25:
```suggestion
    let count;
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,42 @@
...
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        zero = this[i - 1];
+        first = this[i];
```
tetiana-ivanchenko, line 32:
```suggestion
      for (let i = 1; i < this.length; i++) {
        const previousElement = this[i - 1];
        const currentElement = this[i];
```

tetiana-ivanchenko (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,41 @@
...
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  function compareDefault(a, b) {
+    const aString = String(a);
+    const bString = String(b);
+
+    if (aString > bString) {
+      return 1;
+    }
+     if (aString === bString) {
```
tetiana-ivanchenko, line 14:
```suggestion

    if (aString === bString) {
```

tetiana-ivanchenko (changes requested):
fix your code

image

PR #966, opened 2022-12-14
 js_array-method-sort done

EhorSaprykin:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,37 @@
...
  * Implement method Sort
  */
+function stringSortFunction(a, b) {
+  if (String(a) < String(b)) {
+    return -1;
+  } else if (String(a) > String(b)) {
+    return 1;
+  } else {
+    return 0;
+  }
```
polosanya, line 13:
```suggestion
  if (String(a) < String(b)) {
    return -1;
  }
  
  if (String(a) > String(b)) {
    return 1;
  }
  
  return 0;
  }
```
No need for if...else here.
Your code will not go further after return anyway.

src/arrayMethodSort.js
```diff
@@ -3,9 +3,37 @@
...
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = stringSortFunction) {
```
polosanya, line 18:
```suggestion
  [].__proto__.sort2 = function(
    compareFunction = stringSortFunction
  ) {
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,37 @@
...
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const prew = this[i - 1];
```
polosanya, line 25:
```suggestion
        const prev = this[i - 1];
```
Previous -> prev

polosanya (changes requested):

src/arrayMethodSort.js
EhorSaprykin: done

src/arrayMethodSort.js
EhorSaprykin: done

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+    return 1;
+  }
+
+  return 0;
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = stringSortFunction
```
loralevitska, line 19:
```suggestion
  [].__proto__.sort2 = function(
    compareFunction = stringSortFunction
```

loralevitska (approved):
good job!

PR #968, opened 2022-12-14
Initial commit

yaroslav-furhala:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,38 @@
 /**
  * Implement method Sort
  */
+
+function initialSort(a, b) {
+  if (`${a}` > `${b}`) {
```
UnknownAI, line 8: it's better to use the String() constructor, and add two variables to store strings values, to not duplicate the logic and not call the constructor every time

UnknownAI (changes requested):

PR #970, opened 2022-12-17
Solution

Ukrainiane-panda:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,41 @@
...
+          this[i] = prev;
+        }
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+const compareAsStrings = (a, b) => {
```
ivanchuk-yurii, line 29: you can use [localCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

ivanchuk-yurii (approved):

PR #971, opened 2022-12-21
Method Sort was implemented

serge-kontakt:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
polosanya, line 16: No need for else...if, your code will not go further after return anyway.
Let's make your code a bit cleaner
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA < stringB) {
    return -1;
  }
  
  return 0;
```

polosanya (changes requested):
Let's improve your code a bit)

loralevitska (approved):

PR #972, opened 2022-12-21
Solution

AnnyKoval:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
```
Viktor-Kost, line 18: Try not to use `else if `nesting
Here you can apply just `early return`
Like 
```
if (a === b) {
 return
}

if (c === d) {
 return
}
```

Viktor-Kost (changes requested):
GJ!

Need a liitle improvemen

src/arrayMethodSort.js
AnnyKoval: Thank you! Done!

Viktor-Kost (changes requested):
Sorry, have you commit your changes?

AnnyKoval:
image

src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
...
+
+    if (stringA > stringB) {
+      return 1;
+    }
+
+    if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
witflash, line 19: There is no need to check if strings are equal. 
You can do this in simplest way:
```
if (stringA > stringB) {
  return 1;
}

return -1;
```

witflash (approved):
Good job!

PR #973, opened 2022-12-22
initial commit

ihorivna007:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
  */
+function compareFunction(current, next) {
+  const firstValue = String(current);
+  const secondValue = String(next);
+
+  if (firstValue > secondValue) {
+    return 1;
+  } else if (firstValue === secondValue) {
+    return 0;
+  }
```
TarasHoliuk, line 14:
```suggestion
  }
  
  if (firstValue === secondValue) {
    return 0;
  }
```
it's better to avoid `else if` - use common `if` instead if it's possible (when you return value for example, as you do here)

TarasHoliuk (changes requested):
GJ! I've left 1 small comment

tetiana-ivanchenko (approved):

PR #974, opened 2022-12-22
Solution

natalia-maryshko:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
polosanya, line 38: There is no need to use else...if. Your code will not go further after return anyway
```suggestion
    if (stringA > stringB) {
      return 1;
    }
    
    if (stringA < stringB) {
      return -1;
    } 
    
    return 0;
```

polosanya (changes requested):
Let's polish it up)

loralevitska (approved):

PR #980, opened 2022-12-26
Solution

tarasavage:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
polosanya, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = compareAsString) {
```
You can create it and set as default if nothing was passed

src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+        const stringX = String(x);
+        const stringY = String(y);
+
+        if (stringX > stringY) {
+          return 1;
+        } else if (stringX === stringY) {
+          return 0;
+        } else if (stringX < stringY) {
+          return -1;
+        }
```
polosanya, line 22: You can move this logic to the separate function
```suggestion
      comparator = (x, y) => {
        const stringX = String(x);
        const stringY = String(y);

        if (stringX > stringY) {
          return 1;
        }
        
        if (stringX < stringY) {
          return -1;
        }
        
        return 0
```
Also, there is no need for `else if` as the code will not go further after `return` anyway

src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+      };
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
```
polosanya, line 27: Do you really need two nested loops? I think only one is enough

polosanya (changes requested):

tetiana-ivanchenko (approved):

PR #982, opened 2022-12-27
sorting method implemented

MykolaKaradzha:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
```
Viktor-Kost, line 18: try not to use `else if` nesting

You can use early return. Example

```
if (a === b) {
  return 'equal'
}

if (a === c) {
  return 'sth'
}

return 'sb'
```

Viktor-Kost (changes requested):
GJ!
Apply one recommendation

src/arrayMethodSort.js
MykolaKaradzha: got it, thanks

OlEzhA0 (approved):
GJ

PR #991, opened 2023-01-02
Sort

andrieieva-yuliia:


sv-dubov (approved):
Looks good!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+          this[i] = first;
+        }
+      }
+    }
+
+    return this;
   };
 }
 
+const sorted = (a, b) => {
```
Anna-Tkachenko, line 29: function name should contain some action
it's not clear for now what's the main action of this function

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+  const strFirst = a.toString();
+  const strSecond = b.toString();
```
Anna-Tkachenko, line 31: let's try to create more general variable names without using data type names

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+
+  if (strFirst > strSecond) {
+    return 1;
+  } else
+  if (strFirst < strSecond) {
+    return -1;
```
Anna-Tkachenko, line 37:
```suggestion
  } else if (strFirst < strSecond) {
    return -1;
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+  } else {
```
Anna-Tkachenko, line 38: last else is redundant here
just return 0 at the end of the function

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #992, opened 2023-01-03
Solution

Avichez:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  function compareString(a, b) {
+    const firstString = String(a);
+    const secString = String(b);
```
sv-dubov, line 9: You can name it full: like secondString. It's not very long name :) Or better secondStr

sv-dubov (approved):
Looks good!

src/arrayMethodSort.js
Avichez: Got it, thanks) don't like long names and it's difficult to force yourself to write them fully, but I'm trying.

Anna-Tkachenko (approved):
Good job!

PR #994, opened 2023-01-03
Solution

oleh-holovnykh:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+    }
+
+    if (String(a) === String(b)) {
+      return 0;
+    }
+
+    return -1;
+  }) {
+    let prevElem;
+    let currentElem;
```
sv-dubov, line 20: You can name it full: like prevElement etc. It's not very long name :)

sv-dubov (approved):
Looks different but good!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,37 @@
...
+    if (String(a) > String(b) || a === undefined) {
+      return 1;
+    }
+
+    if (String(a) === String(b)) {
+      return 0;
+    }
+
+    return -1;
+  }) {
```
Anna-Tkachenko, line 17: move it to the separately declared function (with it's own name) instead of writing functionality while declaring parameters

Anna-Tkachenko (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+      }
+    } while (count !== 0);
+
+    return this;
   };
-}
+};
+
+function defaultCompare(a, b) {
+  if (String(a) > String(b) || a === undefined) {
```
Anna-Tkachenko, line 32: check if a is undefined here is redundant

Anna-Tkachenko (approved):
Good job!

PR #998, opened 2023-01-04
array sort - completed

GarryRocksman:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,44 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let callback = compareFunction;
+    let prev;
+    let curr;
+    let count;
+
+    const defaultFunction = function(a, b) {
```
sv-dubov, line 13: I think not good naming :)

sv-dubov (approved):
Looks good!

Anna-Tkachenko (changes requested):

src/arrayMethodSort.js
GarryRocksman: Thanks, I will try to code my code more understandable, especially in variable naming ))) I see that I have some trouble with it )))

src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
...
  * Implement method Sort
  */
+
+function compareAsString(a, b) {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
```
Anna-Tkachenko, line 13: else is also redundant here, just move this `if` check next to the first one

Anna-Tkachenko (approved):

src/arrayMethodSort.js
GarryRocksman: Thank you! Fixed it.

PR #999, opened 2023-01-04
Solution

chermak-oleh:


sv-dubov (approved):
Well done!

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringB > stringA) {
+    return -1;
+  } else {
+    return 0;
+  }
```
Anna-Tkachenko, line 16: last else is redundant
just return 0 by default at the end of the function

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
 /**
  * Implement method Sort
  */
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringB > stringA) {
```
Anna-Tkachenko, line 12: else here is also redundant
just create separate if statement after the first one

Anna-Tkachenko (approved):

PR #1001, opened 2023-01-05
solution

NVyrsta:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,37 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
+function compareAsString(a, b) {
+  if (String(a) > String(b)) {
```
andrieieva-yuliia, line 4: You can create variables for strings

andrieieva-yuliia (approved):
GJ 👍

src/arrayMethodSort.js
NVyrsta: thanks for the advice 👍

ArturErmolenko (approved):

PR #1004, opened 2023-01-05
Solution

tania-boiar:


Avichez (approved):
That's a great one solution 👍

src/arrayMethodSort.js
```diff
@@ -5,7 +5,41 @@
...
+      if (stringA > stringB) {
+        return 1;
+      }
+
+      if (stringA < stringB) {
+        return -1;
+      }
+
+      return 0;
+    };
```
Anna-Tkachenko, line 21: will be better to move this function outside of the sort2 method and set it by default while describing function parameters

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1006, opened 2023-01-06
Method sort solution

PolinaSukhorukova:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA < stringB) {
+    return -1;
+  } else {
+    return 0;
```
vanst1, line 15: try to avoid `if else` construction

vanst1 (changes requested):

alexandra-protyanova (approved):

PR #1007, opened 2023-01-06
Solution

Kardash-Oleksandr:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+const compareAsStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
```
vanst1, line 38: try to avoid `if else `construction

vanst1 (changes requested):

src/arrayMethodSort.js
Kardash-Oleksandr: In this case, the use of such a construction is the best option, since replacement by a ternary operator is not possible here, and the use of several "if" operators entails unnecessary checks.

Kardash-Oleksandr: fixed

alexandra-protyanova (approved):

PR #1008, opened 2023-01-06
method sort was implemented

StetsenkoViktoriia:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
Anastasiia-Svintsova, line 39: This else statement is not necessary here
```suggestion
  } 
  
  return -1;
```

Anastasiia-Svintsova (approved):

PR #1010, opened 2023-01-07
solution

imariash86:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,39 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA < stringB) {
+      return -1;
+    } else {
+      return 0;
+    };
+  };
```
vanst1, line 18: try to avoid `if else` construction

vanst1 (changes requested):

vanst1 (approved):
GJ:)

alexandra-protyanova (approved):

PR #1015, opened 2023-01-08
Solution

andrii-andriushchenko:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareAsStr = (a, b) => {
+    const strA = String(a);
+    const strB = String(b);
```
Serghii29, line 9: you can use others variables

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+
+    if (strA > strB) {
+      return 1;
+    } else if (strA < strB) {
+      return -1;
+    } else {
+      return 0;
```
Serghii29, line 16: you can rewrite loop without else

Serghii29 (commented):
Good job👀

src/arrayMethodSort.js
andrii-andriushchenko: fixed, thanks)

src/arrayMethodSort.js
andrii-andriushchenko: renamed)

Denchepornyuk (approved):

PR #1020, opened 2023-01-10
Solution

maasiimka:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+
+      if (condition < 0 && i >= 0) {
+        const temporaryContainer = this[i];
+
+        this[i] = this[i + 1];
+        this[i + 1] = temporaryContainer;
+
+        i -= 2;
+      }
+    }
```
nzrii-kysel, line 36: i'd refactor the function a bit
```suggestion
    for (let i = 0; i < this.length - 1; i++) {
      if (i < 0) {
        continue;
      }
      
      const condition = compareFunction(this[i + 1], this[i]);;

      if (condition < 0) {
        [this[i], this[i + 1]] = [this[i + 1], this[i]];

        i -= 2;
      }
    }
```

nzrii-kysel (approved):
clever approach 👍

src/arrayMethodSort.js
maasiimka: Really, your variant is less complicated. Thanks for review

src/arrayMethodSort.js
```diff
@@ -3,9 +3,38 @@
 /**
  * Implement method Sort
  */
+const defaultCompareFunction = (a, b) => {
+  if (a.toString() < b.toString()) {
+    return -1;
+  }
+
+  if (a.toString() > b.toString()) {
+    return 1;
```
Anna-Tkachenko, line 12: will be better to convert values to string before all checks one time for not repeating your code

Anna-Tkachenko (approved):

PR #1021, opened 2023-01-11
Finish task

U-Ra-All:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,44 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
OlEzhA0, line 19: when you use `if` statements and return something, you can use next syntax

```
if (a > b) {
  return x
}

if (a < b) {
  return y
}

return c
```

OlEzhA0 (approved):
Good job!

src/arrayMethodSort.js
U-Ra-All: Thanks!

PR #1032, opened 2023-01-23
method sort done

kirill-vypirovskyi:


lilia-obushenko (approved):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,33 @@
...
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareStrings = (a, b) => {
+    return (String(a) > String(b))
+      ? 1
+      : (String(a) < String(b))
+        ? -1
+        : 0;
```
Andr1yK, line 12: never use nested ternary operators, it is difficult to read

Andr1yK, line 12: I'm not sure, but it should probably work.

```suggestion
    return (String(a) < String(b))
      ? -1
      : (String(a) > String(b))
```

Andr1yK: But personally, I would rewrite it to ifs :))

Andr1yK (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,37 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareStrings = (a, b) => {
+    if (String(a) > String(b)) {
```
OleksiiBlezniuk, line 8: Or you could save it to constants

OleksiiBlezniuk (approved):
Well done!

PR #1034, opened 2023-01-25
Solution method-sort

NazariiAlieksieiev:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
IvanFesenko, line 17:
```suggestion
    if (stringA > stringB) {
      return 1;
    } 
    
   if (stringA < stringB) {
      return -1;
   } 
    
    return 0;
```

IvanFesenko (approved):
Try do not use else statement where you can do without it

PR #1037, opened 2023-01-28
resolve

Bir3791:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
Volodymyr-Melnyk-Fron10der, line 39: Simplify conditions with `early return pattern`
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA === stringB) {
    return 0;
  } 
  
  return -1;
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+    }
+
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        if (compareFunction(this[i - 1], this[i]) > 0) {
+          [[this[i - 1]], [this[i]]] = [[this[i]], [this[i - 1]]];
```
Volodymyr-Melnyk-Fron10der, line 19: I have no idea how the tests have been passed😅 But I suggest to make in more readable and understandable way
```
const previous = this[i - 1];
const current = this[i];

this[i - 1] = current;
this[i] = previous;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Good, but I suggest yo make some changes)

src/arrayMethodSort.js
Bir3791: [a, b] = [b, a] is the destructuring assignment that swaps the variables a and b.

i chose this method because it short and simple

qoa11a (approved):

PR #1040, opened 2023-01-31
[SOLUTION]

grabynsky:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsStr) {
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 11: we can replace `count` variable with some boolean variable for better readability

what do you think?

etojeDenys (changes requested):
.

src/arrayMethodSort.js
grabynsky: it is possible. I changed

Bohdan-Maksymiuk (approved):

PR #1047, opened 2023-02-03
method sort

bohdan-gavrilyk:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
IvanFesenko, line 16:
```suggestion
  if (stringA > stringB) {
    return 1;
  }
  
  if (stringA === stringB) {
    return 0;
  } 
  
   return -1;
```

IvanFesenko (approved):
Nice work. Keep in mind, less else statement

PR #1050, opened 2023-02-05
Solution

bojkovladislav:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+  const strA = String(a);
+  const strB = String(b);
+
+  if (strA > strB) {
+    return 1;
+  } else if (strA === strB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
SanyaBratashchuk, line 17:
```suggestion
  if (strA > strB) {
    return 1;
  } 
  
  if (strA === strB) {
    return 0;
  } 
    
  return -1;
  
```

SanyaBratashchuk (changes requested):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
...
+    return 0;
+  } 
+    
+  return -1;
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
```
DarkMistyRoom, line 23: you can do the same by passing the default function
`[].__proto__.sort2 = function(
  )`

DarkMistyRoom (approved):

PR #1053, opened 2023-02-07
solution

Roman-Kulchytskyi:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
IvanFesenko, line 16: No need else statement 
```suggestion
  if (stringA > stringB) {
    return 1;
  }

  if (stringA === stringB) {
    return 0;
  }

  return -1;
```

IvanFesenko (changes requested):
Good job. Refactor `compare` function

Viktor-Kost (approved):

PR #1054, opened 2023-02-08
result

YevaDrach:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+  }
+
+  return -1;
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareStrings) {
+    let count;
```
OleziO, line 23: Always initialize your variables.

src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+
+    do {
```
OleziO, line 25: Better to use `for` cycle, then you dont need to create additional variable, and your code gets cleaner and more readible. Also read some info about simplest sorting methods:

1)[ Bubble sort](https://www.geeksforgeeks.org/bubble-sort/)
2) [Selection sort](https://www.geeksforgeeks.org/selection-sort/)

OleziO (approved):

PR #1060, opened 2023-02-09
solution

ivanm2706:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+        let prev = this[i - 1];
+        let curr = this[i]
+
+        if (compareFunction(curr, prev) < 0) {
+          this[i] = prev;
+          this[i - 1] = curr;
+          count++;
+        }
+      }
+    } while (count > 0)
```
etojeDenys, line 36: we can rewrite it using a boolean variable, which will be much more readable

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const strA = String(a);
+    const strB = String(b);
+
+    if (strA > strB) {
+      return 1;
+    } else if (strA === strB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
Volodymyr-Melnyk-Fron10der, line 17: Use early return pattern
```suggestion
    if (strA > strB) {
      return 1;
    }
    
    if (strA === strB) {
      return 0;
    }
    
    return -1;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, looks good, only one comment

etojeDenys (approved):
looks good to me

PR #1061, opened 2023-02-09
Solution

G37ua:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const callbackNull = (a, b) => {
```
witflash, line 7: `callbackNull` looks like strange name :)
It's better to name it smth like `sortDefault`.

witflash (approved):
Good work!

witflash (approved):

PR #1069, opened 2023-02-15
Task done

sergik94:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
+    let callback = compareFunction;
+
+    if (!compareFunction) {
+      callback = function(curr, prev) {
+        return String(curr) < String(prev) ? -1 : 1;
+      };
+    }
+
+    do {
+      step = 0;
```
etojeDenys, line 18: also let's declare here boolean variable not a number

src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let step;
+    let callback = compareFunction;
+
+    if (!compareFunction) {
+      callback = function(curr, prev) {
+        return String(curr) < String(prev) ? -1 : 1;
+      };
```
etojeDenys, line 14: let's rewrite it using the arrow function it would be much more readable
and you need to return `0` if curr === prev

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
Viktor-Kost, line 7: use default value for a function parameter
```suggestion
  [].__proto__.sort2 = function(compareFunction = compareByDefault) {
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
...
-    // write code here
+    let step;
+    let callback = compareFunction;
```
Viktor-Kost, line 9:
```suggestion
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
...
+
+    if (!compareFunction) {
+      callback = (curr, prev) => String(curr) < String(prev) ? -1
+        : String(curr) > String(prev) ? 1 : 0;
+    }
```
Viktor-Kost, line 14: create a separated function `compareByDefault` which you add as a default value (read above)

src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let step;
+    let callback = compareFunction;
+
+    if (!compareFunction) {
+      callback = (curr, prev) => String(curr) < String(prev) ? -1
+        : String(curr) > String(prev) ? 1 : 0;
```
Viktor-Kost, line 13: it's better not use nesting ternary. Just use `if` with `early return`

Example: 
```
if (condition) {
  return value
}

if (condition2) {
  return value2
}

return value3
```

Viktor-Kost (changes requested):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+    return -1;
+  };
+
+  if (String(curr) > String(prev)) {
+    return 1;
+  };
+
+  if (String(curr) === String(prev)) {
+    return 0;
+  };
```
IvanFesenko, line 17:
```suggestion
  return 0;
```

IvanFesenko (approved):
Well done

PR #1070, opened 2023-02-16
sorted

edkido:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
Viktor-Kost, line 16: use early return instead of if/else nesting
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA === stringB) {
    return 0;
  }
  
  return -1;
```

Viktor-Kost (approved):
Nice!
Apply one recommendation

etojeDenys (approved):

PR #1071, opened 2023-02-17
Solution

Ch0p3r84:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
+    let counter;
+
+    do {
+      counter = 0;
```
Volodymyr-Melnyk-Fron10der, line 26: I suggest to use boolean value such a isContinue or smth like that to raise readability of the code

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Great job👍 One remark I left

src/arrayMethodSort.js
Ch0p3r84: Hi Volodymyr,
i am not really sure what you mean here ? Can you tip me on it ?

Volodymyr-Melnyk-Fron10der (approved):
Perfect✅

PR #1075, opened 2023-02-18
Sort method

anton-chornobai:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
SanyaBratashchuk, line 17:
```suggestion
    if (stringA > stringB) {
      return 1;
    } 
    
    if (stringA === stringB) {
      return 0;
    } 
      
    return -1;
    
```

SanyaBratashchuk (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,43 @@
...
+      return 1;
+    }
+    
+    else if (stringA === stringB) {
+      return 0;
+    }
+    
+    else {
+      return -1;
+    }
```
VitaliyBondarenko1982, line 22:
```suggestion
    if (stringA === stringB) {
      return 0;
    }
    
   return -1;
```

VitaliyBondarenko1982 (approved):
Approve, but please check comments about avoiding esle if statement

PR #1080, opened 2023-02-20
Solution

Ssviatt:


src/arrayMethodSort.js
```diff
@@ -5,8 +5,30 @@
...
+            this[j] = temp;
+          }
+        } else {
+          if (`${this[j]}` < `${this[i]}`) {
+            const temp = this[i];
+
+            this[i] = this[j];
+            this[j] = temp;
+          }
+        }
```
Volodymyr-Melnyk-Fron10der, line 26: I suggest just create default compareFunction and use it if there is no callback was passed. Because now you have duplicated code on 14-17 and 21-24 lines
```suggestion
          if (compareFunction(this[j], this[i]) < 0) {
            const temp = this[i];

            this[i] = this[j];
            this[j] = temp;
          }
```

src/arrayMethodSort.js
```diff
@@ -5,8 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const length = this.length;
+
+    for (let i = 0; i < length - 1; i++) {
+      for (let j = i + 1; j < length; j++) {
+        if (compareFunction && typeof compareFunction === 'function') {
+          if (compareFunction(this[j], this[i]) < 0) {
```
Volodymyr-Melnyk-Fron10der, line 13: second argument is always boolean type, but has to be an element of array, check how sort callback works here 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Volodymyr-Melnyk-Fron10der (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,9 +4,30 @@
...
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    if (`${a}` < `${b}`) {
+      return -1;
+    } else if (`${a}` > `${b}`) {
+      return 1;
+    } else {
+      return 0;
+    }
+  }) {
```
Volodymyr-Melnyk-Fron10der, line 15: 1. Move it into separated function
2. use early return pattern
```suggestion
   const defaultCompareFunction = (a, b) => {
    if (`${a}` < `${b}`) {
      return -1;
    } 
    
    if (`${a}` > `${b}`) {
      return 1;
    } 

    return 0;
  }
  
  [].__proto__.sort2 = function(compareFunction = defaultCompareFunction) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Much better now! one more suggestion to make:

ViktorHavryliuk (approved):

PR #1082, opened 2023-02-21
Solution

stanislavcodes:


rashyd-hasratov (approved):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - 1; j++) {
+        const current = this[j];
+        const next = this[j + 1];
+
+        if (compareFunction === undefined) {
```
sergii-nosachenko, line 13: Create a variable that will hold the result of comparing, f.e. `isCurrentBigger` instead of `if...else` construction.
Use ternary.

sergii-nosachenko (changes requested):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - 1; j++) {
+        const current = this[j];
+        const next = this[j + 1];
```
sergii-nosachenko, line 11:
```suggestion
    for (let i = 0; i < this.length - 1; i++) {
      for (let j = i; j < this.length; j++) {
        const current = this[i];
        const next = this[j];
```
Use both loops

sergii-nosachenko (changes requested):

src/arrayMethodSort.js
stanislavcodes: Why? It's just a simple bubble sort
```suggestion
    for (let i = 0; i < this.length; i++) {
      for (let j = 0; j < this.length - i - 1; j++) {
        const current = this[j];
        const next = this[j + 1];
```

sergii-nosachenko (approved):
Good work 👍

PR #1091, opened 2023-02-22
Array Method Sort - Solution

Yevhenii0536:


sergey-lipikhin (approved):

src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+    if (String(prev) > String(current)) {
+      return 1;
+    }
+
+    if (String(prev) < String(current)) {
+      return -1;
+    }
+
+    return 0;
+  };
```
sergii-nosachenko, line 17: For your solution will be enough to return the boolean value `String(prev) > String(current)`. True == 1, False == 0, so your if condition inside `sort2` will work ok.

src/arrayMethodSort.js
```diff
@@ -4,9 +4,38 @@
...
+
+  [].__proto__.sort2 = function(compareFunction = compareStrings) {
+    let count;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const prev = this[i - 1];
```
sergii-nosachenko, line 26: What is `prev`?

sergii-nosachenko (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,9 +4,32 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareStrings = (firstElement, secondElement) => {
+    return String(firstElement) > String(secondElement)
+      ? 1
+      : 0;
```
sergii-nosachenko, line 10:
```suggestion
    return String(firstElement) > String(secondElement);
```

sergii-nosachenko (approved):
Good work, commented

src/arrayMethodSort.js
Yevhenii0536: thanks!

PR #1092, opened 2023-02-22
sort implementation

Yekatery:


andriibezkorovainyi (approved):
👍👍

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    }
+
+    return -1;
+  };
```
sergii-nosachenko, line 18: For your solution will be enough to return the boolean value `stringA > stringB`. True == 1, False == 0, so your if condition inside `sort2` will work ok.

sergii-nosachenko (changes requested):

illiarerun (approved):

PR #1094, opened 2023-02-23
Solution

OlhaKissa:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,34 @@
...
     // write code here
+    let temp = 0;
+    let flag = true;
+
+    while (flag) {
+      flag = false;
+
+      if (compareFunction === undefined) {
+        for (let i = 0; i < this.length - 1; i++) {
+          if (this[i] + '' > this[i + 1] + '') {
```
VitaliyBondarenko1982, line 17: better for reading, as for me
```suggestion
          if (String(this[i]) > String(this[i + 1])) {
```

src/arrayMethodSort.js
```diff
@@ -6,6 +6,34 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    let temp = 0;
+    let flag = true;
+
+    while (flag) {
+      flag = false;
+
+      if (compareFunction === undefined) {
```
VitaliyBondarenko1982, line 15: better to make default compareFunction outside and use as default parameter.

VitaliyBondarenko1982 (changes requested):

PR #1096, opened 2023-02-23
Sort

VadKol:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+    case stringPrevious < stringCurrent:
+      return -1;
+  }
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultCompare) {
+    let count = 0;
```
gretoq, line 21:
```suggestion
    let count;
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringCurrent = String(current);
+
+  switch (true) {
+    case stringPrevious > stringCurrent:
+      return 1;
+
+    case stringPrevious < stringCurrent:
+      return -1;
+  }
+};
```
gretoq, line 17: Why you use switch statement, if  you have 2 variants to return. Use:
  return (...condition)
    ? ...value for true
    : ...value for false.

gretoq (changes requested):
Fix some issues, see the comments

src/arrayMethodSort.js
VadKol: Fixed ✔

src/arrayMethodSort.js
```diff
@@ -3,9 +3,35 @@
 /**
  * Implement method Sort
  */
+const defaultCompare = (previous, current) => {
+  const stringPrevious = String(previous);
+  const stringCurrent = String(current);
+
+  return stringPrevious > stringCurrent
+    ? 1
+    : -1;
```
sergii-nosachenko, line 12: Returning `String(previous) > String(current)` is enough

src/arrayMethodSort.js
```diff
@@ -3,9 +3,35 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultCompare) {
+    let count;
```
sergii-nosachenko, line 17:
```suggestion
    let countSwapped;
```

sergii-nosachenko (changes requested):

src/arrayMethodSort.js
VadKol: Fixed ✔

src/arrayMethodSort.js
VadKol: Fixed ✔

sergii-nosachenko (approved):
Good work 👍

PR #1100, opened 2023-02-24
Solution

Vasyl-Pauk:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
OleziO, line 16: `a.localeCompare(b))` will be the same, you don't need all these conditions.

OleziO (changes requested):

src/arrayMethodSort.js
Vasyl-Pauk: fixed, ty for help :)

AlisaVlasova (approved):

PR #1101, opened 2023-02-24
solution

NikitaStambul:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+        const strB = String(b);
+
+        if (strA === strB) {
+          return 0;
+        } else if (strA < strB) {
+          return -1;
+        } else {
+          return 1;
+        }
+      };
```
KuharchukDmytro, line 21: its better to create separate function for this, and also you shouldnt write nested if else statement
```suggestion
      : (a, b) => {
        const strA = String(a);
        const strB = String(b);

        if (strA === strB) {
          return 0;
        }
        
        if (strA < strB) {
          return -1;
        }
        
        return 1;
      };
```

KuharchukDmytro (approved):
only one issue with if else statement, but otherwise good job!

src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+        const strA = String(a);
+        const strB = String(b);
+
+        if (strA === strB) {
+          return 0;
+        } else if (strA < strB) {
+          return -1;
+        } else {
+          return 1;
+        }
```
sergii-nosachenko, line 20: Returning `String(a) > String(b)` is enough

src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+      };
+
+    let swapped;
+
+    do {
+      swapped = false;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        if (copmareFunc(this[i], this[i + 1]) > 0) {
```
sergii-nosachenko, line 29: Create variables for `this[i]` and `this[i + 1]`

src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+          const temp = this[i];
```
sergii-nosachenko, line 30:
```suggestion
```

sergii-nosachenko (changes requested):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const copmareFunc = typeof compareFunction === 'function'
+      ? compareFunction
+      : (a, b) => {
+        return String(a) > String(b);
+      };
```
sergii-nosachenko, line 12:
```suggestion
      : (a, b) => String(a) > String(b);
```

sergii-nosachenko (approved):
Good work 👍

PR #1106, opened 2023-02-25
Array method Sort v.0.1 - Solution

NedobezhkinOleksandr:


NedobezhkinOleksandr:
Please, find solution for Sort Method task

src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
...
+        } else if (String(a) < String(b)) {
+          return -1;
+        }
+
+        return 0;
+      };
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
```
DarkMistyRoom, line 23: let's think about the case when the array becomes sorted by less amount of iterations. Then the loop will make iterations without any changes. let's use `do-while` instead of the outer `for `and improve the algorithm complexity

DarkMistyRoom (changes requested):

anstsot (approved):
Great job, approved!

PR #1108, opened 2023-02-26
Moroz

morozit:
чому не можна було такий запис ?

```js
function applyCustomSort() {
  [].__proto__.sort2 = function (compareFunction) {
    if (typeof compareFunction !== 'function') {
      compareFunction = function (a, b) {
        if (String(a) > String(b)) {
          return 1;
        } else if (String(a) === String(b)) {
          return 0;
        } else {
          return -1;
        }
      };
    }

          const temp = this[i];

          this[i] = this[j];
          this[j] = temp;
        }
      }
    }

    return this;
  };
}
```

при npm run test
видавало таку помилку 
```js
D:\GitHub\mate-academy\Javascript-Advanced\js_array-method-sort\src\arrayMethodSort.js
  9:7  error  Assignment to function parameter 'compareFunction'  no-param-reassign  
```

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -4,8 +4,29 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  function compareFunction(a, b) {
+    if (String(a) > String(b)) {
+      return 1;
+    } else if (String(a) === String(b)) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
+
+  [].__proto__.sort2 = function (callback = compareFunction) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (callback(this[j], this[i]) < 0) {
+          const temp = this[i];
+
+          this[i] = this[j];
+          this[j] = temp;
+        }
+      }
+    }
+
+    return this;
   };
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7571 @@
 {
   "name": "array-method-sort",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "array-method-sort",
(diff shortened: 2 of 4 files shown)


src/arrayMethodSort.js
```diff
@@ -4,8 +4,29 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (callback(this[j], this[i]) < 0) {
+          const temp = this[i];
+
+          this[i] = this[j];
+          this[j] = temp;
+        }
+      }
+    }
```
anstsot, line 27: you can simplify this code to use only one for a loop because now it iterates too many times (for every symbol it goes `this.length` times)

anstsot (changes requested):
Looks good, but you're using 2 loops instead of 1 `for`.
if you have any questions feel free to ask them in chat

Volodymyr-Melnyk-Fron10der:
> чому не можна було такий запис ?
> 
> ```js
> function applyCustomSort() {
>   [].__proto__.sort2 = function (compareFunction) {
>     if (typeof compareFunction !== 'function') {
>       compareFunction = function (a, b) {
>         if (String(a) > String(b)) {
>           return 1;
>         } else if (String(a) === String(b)) {
>           return 0;
>         } else {
>           return -1;
>         }
>       };
>     }
> 
>     for (let i = 0; i < this.length; i++) {
>       for (let j = i + 1; j < this.length; j++) {
>         if (compareFunction(this[j], this[i]) < 0) {
>           const temp = this[i];
> 
>           this[i] = this[j];
>           this[j] = temp;
>         }
>       }
>     }
> 
>     return this;
>   };
> }
> ```
> 
> при npm run test видавало таку помилку
> 
> ```js
> D:\GitHub\mate-academy\Javascript-Advanced\js_array-method-sort\src\arrayMethodSort.js
>   9:7  error  Assignment to function parameter 'compareFunction'  no-param-reassign  
> ```

Because you tried to reassign compareFunction. Whereas we should write immutable functions which means - not mutate initial state (arguments). In this case you should have just used `default` - `function (compareFunction = someOtherFunction) {`

witflash (approved):
Well done 👍

PR #1113, opened 2023-02-28
Sort created

yevhen1408:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,20 @@
...
+        const previous = this[j - 1];
+        const next = this[j];
+
+        if ((!compareFunction && previous.toString() > next.toString())
+          || (compareFunction && compareFunction(previous, next) > 0)) {
+          this[j] = previous;
+          this[j - 1] = next;
+        }
+      }
+    }
```
DarkMistyRoom, line 19: we shouldn't hardcode the iteration count in the external loop. think about a situation when array became sorted earlier

DarkMistyRoom (changes requested):
good, but let's improve the performance

SerhiiKirik (approved):
Well done 🚀

PR #1114, opened 2023-02-28
create sort

BozhenaBodnar:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,41 @@
...
+
+  if (stringA > stringB) {
+    return 1;
+  }
+
+  if (stringA === stringB) {
+    return 0;
+  }
+
+  return 1;
```
illiarerun, line 41: you only need to return `1` if value is bigger
```suggestion
  return -1;
```

illiarerun (changes requested):

PR #1120, opened 2023-03-03
sort method

RufatRM:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,44 @@
...
+    return 1;
+  }
+
+  if (previousString < currentString) {
+    return -1;
+  }
+
+  if (previousString === currentString) {
+    return 0;
+  }
```
Volodymyr-Melnyk-Fron10der, line 20: redundant condition
```suggestion
  return 0;
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,44 @@
...
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
     // write code here
+    let count;
+
+    do {
+      count = 0;
```
Volodymyr-Melnyk-Fron10der, line 29: it's better to use boolean value, and name it  isResume for example

Volodymyr-Melnyk-Fron10der (changes requested):
Keep it up

polosanya (approved):
Good job!

PR #1121, opened 2023-03-04
solution

serhiilysenko:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,40 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
Volodymyr-Melnyk-Fron10der, line 16: apply early return pattern
```suggestion
  if (stringA > stringB) {
    return 1;
  }
  
  if (stringA === stringB) {
    return 0;
  }
  
  return -1;
```

src/arrayMethodSort.js
```diff
@@ -3,10 +3,40 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(callback = compareByDefault) {
+    let count;
+
+    do {
+      count = 0;
```
Volodymyr-Melnyk-Fron10der, line 24: I suggest to rename it and use as boolean value
```suggestion
      isContinue = false;
```

Volodymyr-Melnyk-Fron10der (changes requested):
JustKeepDoingWhatYoureDoingAleshaDixonGIF

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 19: In the future - Do not add extra files to PR
To add the file to a commit use git add src/.

SerhiiKirik (approved):
Well done 🔥

PR #1126, opened 2023-03-09
Completed task

metja:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
Volodymyr-Melnyk-Fron10der, line 16: use early return pattern
```suggestion
  if (stringA > stringB) {
    return 1;
  } 
  
  if (stringA === stringB) {
    return 0;
  } 
  
  return -1;
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsString) {
+    let count;
+
+    do {
+      count = 0;
```
Volodymyr-Melnyk-Fron10der, line 24: transform it into boolean variable like isResume

Volodymyr-Melnyk-Fron10der (changes requested):
Almost done

src/arrayMethodSort.js
```diff
@@ -3,9 +3,43 @@
...
+    return 1;
+  }
+
+  if (stringA < stringB){
+    return -1;
+  }
+
+  if (stringA === stringB) {
+    return 0;
+  }
```
Volodymyr-Melnyk-Fron10der, line 20: if 2 previous cases were not passed it means that stringA === stringB already
```suggestion
    return 0;
```

Volodymyr-Melnyk-Fron10der (approved):
Ok, just take a look on my comment and do not make this mistake in the future plz👍

PR #1127, opened 2023-03-09
Solution

dudar-y:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
...
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(callback = compareAsStrings) {
+    let count;
+
+    do {
+      count = 0;
```
Volodymyr-Melnyk-Fron10der, line 27: I would rather use boolean type instead of number

Volodymyr-Melnyk-Fron10der (approved):
Looks good👍

PR #1130, opened 2023-03-14
'done'

Hanna-Balabukha:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,40 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
Viktor-Kost, line 17: no needs to use `else if` when you use early return
```suggestion
    if (stringA > stringB) {
      return 1;
    } 
    
    if (stringA === stringB) {
      return 0;
    }
    
    return -1;
```

Viktor-Kost (approved):
Well done
Just apply a recommendation from a comment

PR #1132, opened 2023-03-16
method sort() was created

Pavlo-Petrashevskyi:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
 /**
  * Implement method Sort
  */
+const localeCompare = function(a, b) {
```
DarkMistyRoom, line 6: ok, but it isn't real localeCompare method)

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+    return 0;
+  } else {
+    return -1;
+  }
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = localeCompare) {
```
DarkMistyRoom, line 20:
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b) {
```

DarkMistyRoom: Your solution is good, but I want you to know that it will give the same result

DarkMistyRoom (approved):
Good job!

src/arrayMethodSort.js
Pavlo-Petrashevskyi: Ok, I'll change its name)

src/arrayMethodSort.js
Pavlo-Petrashevskyi: Thank you very much! I haven`t considered comparing them with ">" sign.

PR #1133, opened 2023-03-17
solution

zvir91:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,39 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
Anastasiia-Svintsova, line 17: You don't really need else here 
```suggestion
    }
    
    return -1;
```

Anastasiia-Svintsova (approved):

PR #1134, opened 2023-03-18
Solution

LBondarchuk:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,41 @@
...
+      }
+    } while (count > 0);
+
+    return this;
   };
 }
 
+function compareAsString(a, b) {
+  const stringA = '' + a;
+  const stringB = '' + b;
```
ivbutsykin, line 31: Also you can use `String()` here:
```suggestion
  const stringA = String(a);
  const stringB =  String(b);
```

ivbutsykin (approved):
Good job!

PR #1135, opened 2023-03-20
Implemented sort method

roshirin:


prokivskaaa (approved):
Great job!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
...
+    do {
+      counter = 0;
+
+      for (let i = 0; i < this.length - 1 - shift; i++) {
+        const compareResult = compareFunction(this[i], this[i + 1]);
+
+        if (compareResult > 0) {
+          counter++;
+          [this[i], this[i + 1]] = [this[i + 1], this[i]];
+        }
```
SerhiiKirik, line 22:
```suggestion
      for (let index = 0; index < this.length - 1 - shift; index++) {
        const currentValue = this[index];
        const nextValue = this[index + 1];
        const compareResult = compareFunction(currentValue, nextValue );

        if (compareResult > 0) {
          counter++;
          [currentValue, nextValue] = [nextValue, currentValue];
        }
```

SerhiiKirik (approved):
LGTM 🔥

src/arrayMethodSort.js
roshirin: This solution doesn't sort the array, it only swapes currentValue and nextValue values. But works if we use 
`[this[index], this[index + 1]] = [nextValue, currentValue];`

PR #1139, opened 2023-03-22
solution

Sbavdik:


Kavon-Skela (approved):
You have reached the target, good job. I see a few ways to improve complexity and code style.

You have a loop in another loop. It is O(n^2) algorithm complexity. Try to rewrite it only with the one loop through the input array. But we don't have to keep O(n) complexity in this task, so it is not necessary

Also, your code will be more readable, if you create a new function with the default sort of elements and use it, if it is necessary

SerhiiKirik (changes requested):
Great 🔥 
Totally agree with this [comment](https://github.com/mate-academy/js_array-method-sort/pull/1139#pullrequestreview-1354399406)
Let's try to make your algorithm better

src/arrayMethodSort.js
```diff
@@ -3,9 +3,30 @@
 /**
  * Implement method Sort
  */
+const compareStrings = (firstString, secondString) => {
+  return String(firstString) > String(secondString);
```
SerhiiKirik, line 7: hm, take a look what return callback of sort

SerhiiKirik (changes requested):
Well done 🔥 

PR #1144, opened 2023-03-23
solution

PashaMalyshkin:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  function compareDefault(a, b) {
+    const firstItem = a.toString();
+    const secondItem = b.toString();
+
+    if (firstItem > secondItem) {
+      return 1;
+    } else if (firstItem === secondItem) {
```
BondarenkoDmitriy, line 13: IMHO, else we can't use (from checklist)

BondarenkoDmitriy (approved):
jg

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  function compareDefault(a, b) {
```
liliia-livandovskaya, line 7: it would be better to move this function out of the `compareDefault`. Also,  `a,b` variable names should be replaced with the ones with more context

src/arrayMethodSort.js
liliia-livandovskaya: agree

liliia-livandovskaya (changes requested):
Great job! Lets work on improvement

src/arrayMethodSort.js
```diff
@@ -3,9 +3,38 @@
...
  */
+
+function compareDefault(previousItem, currentItem) {
+  const firstItem = previousItem.toString();
+  const secondItem = currentItem.toString();
+
+  return firstItem > secondItem
+    ? 1
+    : firstItem === secondItem
+      ? 0
```
SerhiiKirik, line 14: try to avoid ternary in ternary

PR #1145, opened 2023-03-23
sort function

PavloReutskiy:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        if (currentSort(this[i - 1], this[i]) > 0) {
+          const transferValue = this[i - 1];
+
+          this[i - 1] = this[i];
+          this[i] = transferValue;
```
roshirin, line 30: You can use destructurization for swapping two elements values without adding an extra variable.

src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+      : compareFunction;
+
+    let count = 0;
+
+    do {
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        if (currentSort(this[i - 1], this[i]) > 0) {
+          const transferValue = this[i - 1];
```
roshirin, line 27: This sorting algorythm can be improved.

First iteration will put the biggest element to the last position, second iteration will put 2nd biggest to 2nd position from the end and so on. So you don't have to iterate through all the array every time.

But it works well without any improvement, this is only a suggestion.

roshirin (approved):
Sorts well, good job!

src/arrayMethodSort.js
PavloReutskiy: it's a great idea! thx a lot! I had to google to understand what really did you mean, but it was very useful!)) I heard about bubble sort, but now I know how to realise it

src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const defaultSort = (elementA, elemenrB) => {
+      return String(elementA) > String(elemenrB)
+        ? 1
+        : String(elementA) === String(elemenrB)
+          ? 0
```
SerhiiKirik, line 12: try to avoid nesting ternary in ternary

SerhiiKirik (changes requested):
Well done 🔥 
Let's improve your code 🧑‍💻

SerhiiKirik (approved):
LGTM 🔥

PR #1146, opened 2023-03-23
Sort

OlegRak86:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
 /**
  * Implement method Sort
  */
+const defaultCallback = (a, b) => {
+  const strA = a + '';
+  const strB = b + '';
```
etojeDenys, line 8:
```suggestion
  const strA = String(a);
  const strB = String(b);
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+
+  if (strA > strB) {
+    return 1;
+  } else if (strA === strB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
etojeDenys, line 16:
```suggestion
  if (strA > strB) {
    return 1;
  } 
  
  if (strA === strB) {
    return 0;
  } 
   
   return -1;
  
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultCallback) {
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 24: we can make this variable a boolean

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultCallback) {
+    let count;
+
+    do {
+      count = false;
```
loralevitska, line 26: You changed variable type but you need also change variable name according new type
[Tips on naming boolean variables](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)

loralevitska (approved):

PR #1150, opened 2023-03-24
Solution

vinnykalisa:


M1k1ta (approved):
Good job!

src/arrayMethodSort.js
```diff
@@ -3,10 +3,34 @@
 /**
  * Implement method Sort
  */
+function compareStrings(previous, current) {
+  const previousString = String(previous);
+  const currentString = String(current);
+
+  return previousString > currentString;
```
SerhiiKirik, line 10: take a look to the doc, what sort return

SerhiiKirik (changes requested):
Great 🔥 

liliia-livandovskaya (approved):

PR #1153, opened 2023-03-25
add custom sort method

prokivskaaa:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,30 @@
 /**
  * Implement method Sort
  */
+const defaultCompare = (prev, curr) => String(prev) > String(curr) ? 1 : -1;
```
SerhiiKirik, line 6: what should return callback if values is equal?

SerhiiKirik (approved):
Great 🔥

PR #1156, opened 2023-03-25
solution

mzaloha:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(callback = functionCompare) {
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 26: we can use here a boolean variable

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+  }
+
+  return -1;
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(callback = functionCompare) {
+    let count = false;
```
sTorba24, line 23:
```suggestion
    let count
```

sTorba24 (approved):
Good job!

PR #1159, opened 2023-03-27
Solution

Dmytro-Kretov:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
+
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
```
DarkMistyRoom, line 20:
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b)) {
```

DarkMistyRoom: it gives the same result

DarkMistyRoom (approved):
Good job!

PR #1162, opened 2023-03-27
Solution

Zibi95:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    this.forEach(() => {
+      for (let i = 0; i < this.length - 1; i++) {
+        const a = this[i];
+        const b = this[i + 1];
+
+        const compare = arguments.length === 0
```
Radoslaw-Czerniawski, line 13: This could have been  `compareFunction === undefined` rather than based on length of arguments property

PR #1164, opened 2023-03-28
develop

DanielSorokowski:


solaryasha (changes requested):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,27 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (compareFunction === undefined) {
+          if (String(this[i]) > String(this[j])) {
+            rememberValue = this[i];
+            this[i] = this[j];
+            this[j] = rememberValue;
+          }
+        } else {
+          if (compareFunction(this[i], this[j]) > 0) {
```
solaryasha, line 19:
```suggestion
        const condition = compareFunction
         ? compareFunction(this[i], this[j]) > 0
         :  String(this[i]) > String(this[j])
         
          if (condition) {
            /* your code */
          }
```

solaryasha (changes requested):

src/arrayMethodSort.js
DanielSorokowski: I understand the changes, but I would never come up with that solution on my own with my current javascript level is it only an efficient fix or are there other reasons why my previous answer was bad?

solaryasha: It's all about [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). You should follow it.
I doesn't mean that your code is bad, it's more about some improvments

solaryasha (approved):
👍

src/arrayMethodSort.js
solaryasha: @DanielSorokowski

PR #1166, opened 2023-03-29
1st commit

ihor-halkin:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const prevElement = String(a);
+    const currentElement = String(b);
+
+    if (currentElement > prevElement) {
+      return -1;
+    } else if (currentElement < prevElement) {
+      return 1;
+    } else {
+      return 0;
+    }
```
polosanya, line 17:
```suggestion
    if (currentElement > prevElement) {
      return -1;
    }
    
    if (currentElement < prevElement) {
      return 1;
    }
    
    return 0;
```
There is no need to use else...if here, as your code will not go further after return anyway

polosanya (changes requested):
Good job! Let's improve your code a bit)

PR #1177, opened 2023-04-05
Solution

NadiiaR:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsString) {
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 11: we can make this variable a boolean

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+
+      for (let i = 1; i < this.length; i++) {
+        const prev = this[i - 1];
+        const current = this[i];
+
+        if (compareFunction(prev, current) > 0) {
+          count++;
```
etojeDenys, line 18: set here true

src/arrayMethodSort.js
```diff
@@ -4,9 +4,40 @@
...
+          this[i - 1] = current;
+          this[i] = prev;
+        }
+      }
+    } while (count > 0);
```
etojeDenys, line 23: and check here if this var is true

etojeDenys (changes requested):

TarasHoliuk (approved):
LGTM :+1:

PR #1181, opened 2023-04-06
Solution

yurii-shkrobut-m:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,26 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) > String(b)) {
```
etojeDenys, line 8: if a === b you need to return 0

src/arrayMethodSort.js
```diff
@@ -4,8 +4,26 @@
...
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 12: we can make this variable a boolean it will improve readability

etojeDenys (changes requested):

src/arrayMethodSort.js
SerhiiKirik: Agree
Take a look what is callback of sort return

SerhiiKirik: Split this to the separate fucntion

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareAsStrings = (a, b) => String(a) > String(b);
```
SerhiiKirik, line 7: So, what is a return callback of sort??

SerhiiKirik (changes requested):
Well done 🔥 
Let's make some improvements 🧑‍💻

anstsot (approved):
Good job, approved!

PR #1183, opened 2023-04-08
solution

Dka888:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,51 @@
...
+          const a = this[i];
+          const b = this[i + 1];
+
+          if (compareFunction(a, b) > 0) {
+            this[i] = b;
+            this[i + 1] = a;
+          }
+        }
+        lengthOfArray--;
+      }
```
d-wasilewski, line 49: a lot of duplication, can you try to solve it in  a more concise way?

d-wasilewski (changes requested):

PR #1185, opened 2023-04-09
array method sort solution

RomanTsvindin:


src/arrayMethodSort.js
```diff
@@ -5,8 +5,26 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
polosanya, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = defaultCompareFunction) {
```
You can create a default sorting function and set it as a default value of compareFunction.
It will help to avoid extra checks and make your code cleaner

src/arrayMethodSort.js
```diff
@@ -5,8 +5,26 @@
...
-    // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      let minIndex = i;
+
+      for (let j = i + 1; j < this.length; j++) {
```
polosanya, line 11: Avoid using nested loops. It's bad for performance and actually makes no sense here. 
`this[i]` - current element
`this[i + 1]` - next element

polosanya (changes requested):
Good job! Let's improve your solution a bit)

PR #1188, opened 2023-04-10
Solution

AnastassiaH:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,41 @@
...
+        for (let i = 0; i < this.length - 1; i++) {
+          const a = this[i];
+          const b = this[i + 1];
+
+          if (compareFunction.bind(this)(a, b) > 0) {
+            this[i] = b;
+            this[i + 1] = a;
+          };
+
+          if ((compareFunction.bind(this)(a, b) <= 0)) {
```
Viktor-Kost, line 21: This if condition is redundant, as the values are already set correctly in the previous if condition.

src/arrayMethodSort.js
```diff
@@ -6,6 +6,41 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
Viktor-Kost, line 7: create default function to use instead of `if` condition
So this function will invoke the same as you did with `compareFunction `

```
  const compareFunctionByDefault = (a, b) => {
    const stringA = String(a);
    const stringB = String(b);

    if (stringA > stringB) {
      return 1;
    } 
    
    if (stringA === stringB) {
      return 0;
    } 
    
    return -1;
  };

```

```suggestion
  [].__proto__.sort2 = function(compareFunction = compareFunctionByDefault) {
```

src/arrayMethodSort.js
```diff
@@ -6,6 +6,36 @@
...
```
YegorKorenkov, line 7: Previous review comment not fixed, please add this additional function and use it like default value of compareFunction argument

YegorKorenkov (changes requested):

tetiana-ivanchenko (approved):

PR #1190, opened 2023-04-10
Solution

NYasnaya:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
```
SerhiiKirik, line 18: try to avoid `else / if` constractions
```suggestion
    if (stringA > stringB) {
      return 1;
    } 
    
   if (stringA === stringB) {
      return 0;
   } 
  
    return -1;
  };
```

SerhiiKirik (approved):
Well done 🚀

PR #1199, opened 2023-04-16
solution

mariuszgit:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,46 @@
...
+    if (compareFunction === undefined) {
+      return this;
+    }
+
+    do {
+      count = 0;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        const a = this[i];
+        const b = this[i + 1];
```
DorotaLeniecDev, line 35: For readibility you could name them like:
```suggestion
        const current = this[i];
        const next = this[i + 1];
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,46 @@
...
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsString) {
+    let count;
+
+    if (compareFunction === undefined) {
+      return this;
+    }
```
DorotaLeniecDev, line 28: You could improve error handling as when **compareFunction === undefined** it will use the default **compareAsString**
```suggestion
    if (typeof compareFunction !== 'function') {
      throw new Error('compareFunction is not valid')
    }
```

DorotaLeniecDev (changes requested):
Good solution but small changes needed

DorotaLeniecDev (approved):
Good job!

PR #1201, opened 2023-04-18
Solution

oksana-skuba:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
+  const strB = String(b);
+
+  if (strA > strB) {
+    return 1;
+  } else if (strA === strB) {
+    return 0;
+  } else {
+    return -1;
+  }
+};
```
NataliiaSolotva, line 18: You can use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

NataliiaSolotva (approved):
Your work is approved! Well done!

PR #1204, opened 2023-04-19
Develop

DefaultGitUser:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,37 @@
...
-    // write code here
+  function compareAsStrings(a, b) {
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else {
+      return -1;
+    }
```
IvanFesenko, line 15: Do not use else statment
what will return if values are equal?

IvanFesenko (changes requested):
Small fix required

etojeDenys (approved):
lgtm

PR #1209, opened 2023-04-23
Solution method sort

vlad-liakh:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareFunctionWithString) {
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 11: make this var boolean to improve readability

etojeDenys (changes requested):

mykhalenych (approved):

PR #1213, opened 2023-04-29
solution

LiliiaDanylenko:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,35 @@
 /**
  * Implement method Sort
  */
+const compareToString = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  }
```
Volodymyr-Melnyk-Fron10der, line 12: compareToString  should return 0 and -1 as well

src/arrayMethodSort.js
```diff
@@ -3,9 +3,35 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareToString) {
+    let count;
+
+    do {
+      count = 0;
```
Volodymyr-Melnyk-Fron10der, line 20: transform to the boolean value

Volodymyr-Melnyk-Fron10der (changes requested):

Volodymyr-Melnyk-Fron10der (approved):

PR #1214, opened 2023-04-29
Solution

andrijdudar:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,35 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareAsStrings = (a, b) => {
+    if (String(a) > String(b)) {
+      return 1;
+    } else if (String(a) === String(b)) {
+      return 0;
+    } else {
+      return -1;
+    };
```
polosanya, line 14:
```suggestion
    if (String(a) > String(b)) {
      return 1;
    }
    
    if (String(a) === String(b)) {
      return 0;
    }
    
    return -1;
```
There is no need to use "else...if", the code will not go after "return" anyway

polosanya (changes requested):
Almost perfect! Let's improve your code a bit)

SanyaBratashchuk (approved):

PR #1218, opened 2023-05-03
solution

generatedman:


ruslanpashkov (approved):
🔥

src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
...
  * Implement method Sort
  */
+
+const compareStrings = (a, b) => {
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
```
33Orange, line 13: Try to avoid `else if` and `else` statements.
```
if (stringA > stringB) {
    return 1;
} 
if (stringA === stringB) {
   return 0;
}
return -1;
```

33Orange (approved):
Generally everything is fine. Well done!

PR #1222, opened 2023-05-03
Add solution

ruslanpashkov:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compare) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - 1; j++) {
```
silchyn, line 9: the second loop can be significantly optimized; think about it

silchyn (changes requested):

PR #1227, opened 2023-05-06
Sort Method is ready!

vanya-kalyenichenko:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+      count = 0;
+
+      for (let i = 1; i < this.length; i++) {
+        const prev = this[i - 1];
+        const current = this[i];
+
+        if (compareFunction(prev, current) > 0) {
+          count++;
+          this[i - 1] = current;
+          this[1] = prev;
```
SanyaBratashchuk, line 33: Incorrect index used for assignment, should be 'this[i] = prev;' instead of 'this[1] = prev;'

SanyaBratashchuk (changes requested):

ivbutsykin (approved):
Good job!

PR #1238, opened 2023-05-12
SortMethod

Dmytro-Tymchenko:


src/arrayMethodSort.js
```diff
@@ -1,12 +1,28 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
+const compareAsSrting = (a, b) =>
```
etojeDenys, line 3: Typo in the function name 'compareAsSrting', should be 'compareAsString'

etojeDenys (approved):

PR #1248, opened 2023-05-16
Solution

VKonoplianyi:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,32 @@
...
+      compare = (a, b) => {
+        if (a.toString() > b.toString()) {
+          return 1;
+        } else if (a.toString() < b.toString()) {
+          return -1;
+        }
+
+        return 0;
+      };
+    }
```
anstsot, line 20: create separate function outside the `sort2` method and set is as default for `compareFunction` argument

src/arrayMethodSort.js
```diff
@@ -5,7 +5,32 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        if (compare(this[j], this[j + 1]) > 0) {
+          const temp = this[j];
+
+          this[j] = this[j + 1];
+          this[j + 1] = temp;
+        }
+      }
+    }
```
anstsot, line 31: the logic can be simplified. Try to use `do ... while` instead of using 2 loops

anstsot (changes requested):
It’s almost done! Let’s improve a little

SanyaBratashchuk (approved):

PR #1249, opened 2023-05-17
sort method

pos1bl:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,40 @@
 /**
  * Implement method Sort
  */
+
+const defailtCompare = (a, b) => {
```
volodymyr-soltys97, line 7: Typo in the function name 'defailtCompare', it should be 'defaultCompare'

volodymyr-soltys97 (approved):
Great work 🚀 
But I think it would be better to correct the error in the function name

PR #1253, opened 2023-05-18
add functionality to applyCustomSort

cephy77:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,47 @@
...
+    if (this.length <= 1) {
+      return this;
+    }
+
+    let counter = 0;
+
+    while (counter < this.length) {
+      let forward = 1;
+
+      for (; forward < this.length; forward++) {
```
33Orange, line 17: Why are you creating a variable above the loop?

src/arrayMethodSort.js
```diff
@@ -4,9 +4,47 @@
...
+    if (forwardChar > previousChar) {
+      return 1;
+    }
+
+    if (previousChar > forwardChar) {
+      return -1;
+    }
+
+    return 0;
+  }
```
33Orange, line 47: Use more descriptive name. This function compare strings.
`forwardChar`, `previousChar` What is forward and prev? 

This function compares two values. Of the name should not be tied to something

33Orange (changes requested):
Good job.

A few moment to fix

dinaminahi (approved):

PR #1254, opened 2023-05-18
made task js array-method-sort

vladyslav-lapshyn:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const stringA = String(valueA);
+  const stringB = String(valueB);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA < stringB) {
+    return -1;
+  } else {
+    return 0;
+  }
```
DidenkoDenis306, line 16: Try to avoid else if and else statements.

if (stringA > stringB) {
    return 1;
} 
if (stringA === stringB) {
   return 0;
}
return -1;

DidenkoDenis306 (changes requested):
Good job, but woulde be better to fix one moment

src/arrayMethodSort.js
dinaminahi: @DidenkoDenis306 is right, let's fix this moment and it will be ready to go

dinaminahi (changes requested):
Good job! Let's fix the moment with `if-else`

src/arrayMethodSort.js
vladyslav-lapshyn: @DidenkoDenis306 thanks, my mistake

33Orange (approved):
Well done ! 🔥

PR #1255, opened 2023-05-18
Solution

bohdan-mykhailenko:


src/arrayMethodSort.test.js
```diff
@@ -43,7 +43,7 @@ test(`sort strings: ['c','b','a','d'].sort2()`, () => {
 
 test(`sort cyrillic letters: ['ёж','як','аист'].sort2()`, () => {
   expect(['ёж', 'як', 'аист'].sort2())
-    .toEqual(['аист', 'як', 'ёж']);
+    .toEqual(['аист', 'ёж', 'як']);
```
dinaminahi, line 46: You can't change the test

src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(callback) {
+    const compareFunction = callback
+    || ((a, b) => String(a).localeCompare(String(b)));
```
dinaminahi, line 9: Instead of  using || you can give this function as the default value for `callback` argument

dinaminahi, line 9: The reason the test is failing is this default compare function. 
It is better to use something like this ` (a, b) => a.toString() > b.toString()`

dinaminahi (changes requested):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,26 @@
 /**
  * Implement method Sort
  */
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(callback = (a, b) => String(a) > (String(b))) {
```
silchyn, line 8:
```suggestion
  [].__proto__.sort2 = function(comparator = (a, b) => String(a) > String(b)) {
```

silchyn (changes requested):

33Orange (changes requested):
check @silchyn comment

silchyn (approved):
that's cool 👍
I meant the faster bubble sort variant, but that fits

PR #1259, opened 2023-05-20
array sort add solution

YeMyr:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,29 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => String(a) < String(b)
+  ) {
+    let transposition = 0;
```
silchyn, line 10: what does transposition mean here?

src/arrayMethodSort.js
```diff
@@ -4,8 +4,29 @@
...
+
+    do {
+      transposition = 0;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        const currentItem = this[i];
+        const nextItem = this[i + 1];
+        const compareMethod = compareFunction(nextItem, currentItem);
```
silchyn, line 18: is it really method?

src/arrayMethodSort.js
```diff
@@ -4,8 +4,29 @@
...
+
+        if (compareMethod < 0 || compareMethod === true) {
+          this[i] = nextItem;
+          this[i + 1] = currentItem;
+          transposition++;
+          break;
```
silchyn, line 24: why?

src/arrayMethodSort.js
```diff
@@ -4,8 +4,29 @@
...
+        }
+      }
+    } while (transposition > 0);
```
silchyn, line 27: think about more optimized solution

silchyn (changes requested):

src/arrayMethodSort.js
YeMyr: tried to optimize solution and fixed mistakes, please, take a look
Thanks

src/arrayMethodSort.js
YeMyr: artifact from initial version
removed it

src/arrayMethodSort.js
YeMyr: renamed it to `itemsComparetion`

src/arrayMethodSort.js
YeMyr: renamed it to `countChanges`

src/arrayMethodSort.js
```diff
@@ -4,8 +4,30 @@
...
+        const nextItem = this[i + 1];
+        const itemsComparetion = compareFunction(nextItem, currentItem);
+
+        if (itemsComparetion < 0) {
+          this[i] = nextItem;
+          this[i + 1] = currentItem;
+          countChanges++;
+        }
+      }
+    } while (countChanges > 0);
```
silchyn, line 28: counting changes and going through the whole array each time is not the best way to implement bubble sort

silchyn (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+  ) {
+    let swapped;
+
+    for (let i = 0; i < this.length - 1; i++) {
+      swapped = false;
+
+      for (let j = 0; j < this.length - i - 1; j++) {
+        const currentItem = this[j];
+        const nextItem = this[j + 1];
+        const itemsComparetion = compareFunction(nextItem, currentItem);
```
silchyn, line 20:
```suggestion
        const comparison = compareFunction(nextItem, currentItem);
```

silchyn (approved):

33Orange (approved):

PR #1260, opened 2023-05-21
Solution

Mykola-Maik:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const stringA = String(a);
+    const stringB = String(b);
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
silchyn, line 17:
```suggestion
    if (stringA > stringB) {
      return 1;
    }
    
    if (stringA === stringB) {
      return 0;
    }
    
    return -1;
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const compareAsStrings = (a, b) => {
```
silchyn, line 7:
```suggestion
  const comparator = (a, b) => {
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+        const previous = this[i - 1];
+        const current = this[i];
+
+        if (compareFunction(previous, current) > 0) {
+          counter++;
+          this[i - 1] = current;
+          this[i] = previous;
+        }
+      }
+    } while (counter > 0);
```
silchyn, line 36: think about a more optimized solution

silchyn (changes requested):

src/arrayMethodSort.js
silchyn: here is useful information on this topic
https://www.geeksforgeeks.org/bubble-sort/

dinaminahi: Please, review @silchyn comment, instead of counting changes and going through the whole array each time you can use boolean variable and stop the algorithm if the inner loop didn’t cause any swap.

dinaminahi (changes requested):

PR #1267, opened 2023-05-24
code

liza-strykharchuk:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,30 @@
 /**
  * Implement method Sort
  */
+const compareAsStrings = function(a, b) {
+  return String(a) > String(b);
```
volodymyr-soltys97, line 7: The compareAsStrings function should return a numeric value, not a boolean. Change the return statement to 'return String(a).localeCompare(String(b));'

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/arrayMethodSort.js
liza-strykharchuk: i dkn why, but it returns erorr in sort cyrillic letters: ['ёж','як','аист'].sort2()   
      "аист",
    -   "як",
        "ёж",
    +   "як",

volodymyr-soltys97 (approved):

PR #1274, opened 2023-05-30
add solution

dimarogkov:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,43 @@
-'use strict';
+/* eslint-disable space-before-function-paren */
+/* eslint-disable quotes */
+"use strict";
 
 /**
  * Implement method Sort
```
OleziO, line 6: Remove all comments from the code.

src/arrayMethodSort.js
```diff
@@ -1,11 +1,38 @@
...
-    // write code here
+  const defaultFunction = (a, b) => {
+    const strA = String(a);
+    const strB = String(b);
+
+    if (strA > strB) {
+      return 1;
+    } else if (strA === strB) {
+      return 0;
+    }
```
IvanFesenko, line 12: Else statement is redudant
```suggestion
    if (strA > strB) {
      return 1;
    } 
    
    if (strA === strB) {
      return 0;
    }
```

IvanFesenko (approved):

PR #1278, opened 2023-05-31
Solution

reseneweb:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,39 @@
...
+  const stringA = String(a);
+  const stringB = String(b);
+
+  if (stringA > stringB) {
+    return 1;
+  } else if (stringA === stringB) {
+    return 0;
+  } else {
+    return -1;
+  }
```
polosanya, line 39:
```suggestion
  if (stringA > stringB) {
    return 1;
  }
  
  if (stringA === stringB) {
    return 0;
  }
  
  return -1;
```
There is no need to use `else...if` construction. Code will not go further after `return` anyway

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #1285, opened 2023-06-07
solution

AndriiYelieva:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (
+          compareFunction
+            ? compareFunction(this[i], this[j]) > 0
+            : String(this[i]) > String(this[j])
```
Viktor-Kost, line 13: It's better to create a default comparing function outside applyCustomSort function and set it as a default value

Example: 
```
const compareAsString = (a, b) => {
...
}

...

...
```

Viktor-Kost (changes requested):

src/arrayMethodSort.js
```diff
@@ -1,11 +1,38 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 21: we can use a boolean variable here

src/arrayMethodSort.js
```diff
@@ -1,11 +1,38 @@
...
+
+      for (let i = 1; i < this.length; i++) {
+        const prev = this[i - 1];
+        const current = this[i];
+
+        if (compareFunction(prev, current) > 0) {
+          count++;
```
etojeDenys, line 28: set here to true

src/arrayMethodSort.js
```diff
@@ -1,11 +1,38 @@
...
+          this[i - 1] = current;
+          this[i] = prev;
+        }
+      }
+    } while (count > 0);
```
etojeDenys, line 33: and check if true here

etojeDenys (changes requested):
it will improve readability

src/arrayMethodSort.js
```diff
@@ -1,11 +1,24 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    for (let i = 0; i < this.length - 1; i++) {
```
IvanFesenko, line 6: Create separate `sortString` that will be used if no `compareFunction` 

```suggestion
  [].__proto__.sort2 = function(compareFunction  = sortString) {
    // write code here
    for (let i = 0; i < this.length - 1; i++) {
```

src/arrayMethodSort.js
```diff
@@ -1,11 +1,24 @@
...
+      for (let j = 0; j < this.length - i - 1; j++) {
+        const shouldSwap = compareFunction
+          ? compareFunction(this[j], this[j + 1]) > 0
+          : String(this[j]) > String(this[j + 1]);
```
IvanFesenko, line 10:
```suggestion
        const shouldSwap = compareFunction(this[j], this[j + 1]) > 0;
```

IvanFesenko (changes requested):

PR #1287, opened 2023-06-09
Solution

andrii-rykal:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
 /**
  * Implement method Sort
  */
```
OleziO, line 5: Remove all comments

src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
...
+
+    if (itemA > itemB) {
+      return 1;
+    }
+
+    if (itemA === itemB) {
+      return 0;
+    }
+
+    return -1;
```
OleziO, line 20: Replace with `return itemA > itemB ? 1 : -1;`

OleziO (changes requested):

src/arrayMethodSort.js
OleziO: Not resolved.

OleziO (changes requested):

VitaliyBondarenko1982 (approved):

PR #1292, opened 2023-06-13
add task solution

KandaskalovOleg:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = function(a, b) {
+      return String(a) < String(b) ? -1
+        : (String(a) > String(b) ? 1 : 0);
+    }
```
anstsot, line 11: try to create separate function for default `compareFunction` 
and do not use 2 ternary operators in a row because it adds complexity to read

anstsot (changes requested):
Almost done :+1:
Let's improve a little

src/arrayMethodSort.js
```diff
@@ -4,9 +4,31 @@
...
+          [this[i], this[i + 1]] = [this[i + 1], this[i]];
+        }
+      }
+    }
+
+    return this;
   };
+
+  function defaultFunction(a, b) {
+    if (String(a) < String(b)) {
```
YegorKorenkov, line 20: On row 10 yout condition check if result of default compare function > 0, then perform some action. So check wich return -1 and 0 redundant, you can shorten them:
if (String(a) > String(b)) {
      return 1;
 }
 
 return 0

YegorKorenkov (approved):
Good job!

PR #1296, opened 2023-06-15
add task solution

kbekher:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,34 @@
...
+          }
+        }
+
+        if (!compareFunction) {
+          if (`${prev}` > `${current}`) {
+            count = true;
+            this[i - 1] = current;
+            this[i] = prev;
+          }
+        }
```
IvanFesenko, line 31: Create separate `compareFunction` that will be used in case when no compare function will used

src/arrayMethodSort.js
```diff
@@ -5,7 +5,34 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
IvanFesenko, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = defaultStringSorting) {
```

IvanFesenko (approved):

PR #1301, opened 2023-06-17
solution

cebanoleksandr:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const compareDefault = (a, b) => Number(a.toString() > b.toString());
+
+    const compare = typeof compareFunction !== 'function'
+      ? compareDefault
+      : compareFunction;
```
adv555, line 12:
```suggestion
const compare = compareFunction || compareDefault;
```
operator || will return the first truthy value. looks more convenient to me)

adv555 (approved):
Interesting solution. led to ideas on how to improve my own code 😊  Approve 👍

PR #1306, opened 2023-06-18
add solution hard

prostonadobobb:
i don`t now how i do, but i do

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -6,6 +6,35 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    let finished = false;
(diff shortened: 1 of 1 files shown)


src/arrayMethodSort.js
```diff
@@ -6,6 +6,35 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    let finished = false;
+    let func = compareFunction;
```
volodymyr-soltys97, line 10: Instead of creating a separate variable 'func', you can directly use 'compareFunction' and assign a default implementation if not provided.

volodymyr-soltys97 (approved):
Good job 👍

PR #1315, opened 2023-06-20
Solution

olena-ole:


anstsot (approved):
The code is a little bit overcomplicated. There is a solution to make it simpler. But it works so I'm approving:sparkles:

PR #1318, opened 2023-06-21
array-method-sort solution

NataliGru:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,24 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareF) {
```
vladmakohon, line 7: Good decision to put the default comparing function right to the parameter. I didn't think of that!

vladmakohon (approved):
Looks good to me!😄

src/arrayMethodSort.js
NataliGru: Thanks)

tetiana-ivanchenko (approved):

PR #1325, opened 2023-06-26
Sort method clone

ValeraViachalo:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,37 @@
...
+      sortFunction = function(firstValue, secondValue) {
+        if (String(firstValue) > String(secondValue)) {
+          return 1;
+        } else if (String(firstValue) < String(secondValue)) {
+          return -1;
+        }
+
+        return 0;
+      };
+    }
```
Viktor-Kost, line 20: it's better move this logic to an own function and use as default function parameter 

```
const compareAsStrings = (a, b) => {
  const stringA = String(a);
  const stringB = String(b);
  
  if (stringA > stringB) {
    return 1;
  }
  
  if (stringA === stringB) {
    return 0;
  }
    
  return -1;
};
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,37 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
Viktor-Kost, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
```

IvanFesenko (approved):

PR #1326, opened 2023-06-26
solution

venher-v:


src/arrayMethodSort.js
```diff
@@ -3,10 +3,35 @@
 /**
  * Implement method Sort
  */
+
+const compareStringFunction = (x, y) => {
+  const stringX = x.toString();
+  const stringY = y.toString();
+
+  return stringX > stringY ? 1 : stringX < stringY ? -1 : 0;
```
etojeDenys, line 11: better to use `localeCompare`

etojeDenys (changes requested):

loralevitska (approved):
Good job!

src/arrayMethodSort.js
loralevitska: just commit called `the proposed option does not work`, it is not suitable. Describe the code you wrote, you can also always write to the chat to solve the problem and find out why the code does not work

PR #1334, opened 2023-07-05
Solution

Luk2asz:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
...
+    do {
+      counter = 0;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        const aString = String(this[i]);
+        const bString = String(this[i + 1]);
+        const a = this[i];
+        const b = this[i + 1];
+
+        if (compareFunction === undefined && aString > bString) {
```
darokrk, line 19: Think you can modify this `if else` statement into one to be:

`if (compareFunction === undefined && aString > bString || compareFunction !== undefined && compareFunction(a, b) > 0)`

They doing the same operations

darokrk (changes requested):
Left one suggestion to fix, and think we are good to go 👍🏽

src/arrayMethodSort.js
darokrk: `if ((compareFunction === undefined && aString > bString) || (compareFunction !== undefined && compareFunction(a, b) > 0))
`

how about this one?

Luk2asz: still the same

DorotaLeniecDev (changes requested):
Make use of default function parameters

src/arrayMethodSort.js
```diff
@@ -4,9 +4,27 @@
...
+    let counter;
+
+    do {
+      counter = 0;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        const a = this[i];
+        const b = this[i + 1];
+
+        if (compareFunction !== undefined && compareFunction(a, b) > 0) {
```
DorotaLeniecDev, line 19: Using default function parameter we dont need to check if function is not undefined
```suggestion
        if (compareFunction(a, b) > 0) {
```

DorotaLeniecDev (approved):
Good job, one more suggestion but I am approving

PR #1335, opened 2023-07-06
solution

Constantin-Haranskiy:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const defaultCallback = (a, b) => String(a) >= String(b);
```
SanyaBratashchuk, line 7: it is better to return number to avoid implicit comparison on 14 line

SanyaBratashchuk (approved):

PR #1340, opened 2023-07-07
Solution

AytajBabayeva:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,30 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    const array = this;
```
Anastasiia-Svintsova, line 9: Directly reassigning 'this' to another variable is not a good practice. Use `this` directly in this task

src/arrayMethodSort.js
```diff
@@ -6,6 +6,30 @@
...
+    const length = array.length;
+
+    for (let i = 0; i < length - 1; i++) {
+      for (let j = 0; j < length - 1 - i; j++) {
+        if (compareFunction && typeof compareFunction === 'function') {
```
Anastasiia-Svintsova, line 14: compareFunction should always exist, so this checking is redundant. Just add a default value for the compareFunction in arguments 
```
function(compareFunction = defaultCompareFn) { 
 ...

Anastasiia-Svintsova (changes requested):

PR #1342, opened 2023-07-08
my solution

misha-shevchuk:
First, we are checking if the array contains more than two items. If not - we have to return the original array. Otherwise, we are comparing every item from the array with all the other items using two loops, one inside another. The comparison is made by using a custom function, which converted every item to a string and compared them. If the current item is greater than the next item we need to swap them.

Maybe the complicity of this algorithm is too high for this kind of task(because of a loop inside the loop), but it was the only working solution I`ve come up with and which I totally understand by now.

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -4,9 +4,31 @@
  * Implement method Sort
  */
 function applyCustomSort() {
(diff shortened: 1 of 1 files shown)


volodymyr-soltys97 (approved):
Nice work 🔥

PR #1345, opened 2023-07-09
solution

Hladkyi03:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
+        for (let j = 0; j < this.length - i - 1; j++) {
+          if (String(this[j]) > String(this[j + 1])) {
+            const temporary = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temporary;
+          }
+        }
+      }
+    }
```
VitaliyBodnar, line 30: let's improve your solution )
 - DRY - don't repeat yourself 
 - you have a lot of code duplications
 - for fixing that you can set for the compareFunction a default value, for example 
 
Screenshot 2023-07-09 at 16 28 49

VitaliyBodnar (changes requested):
Well done )

VitaliyBodnar (changes requested):
Request changes

src/arrayMethodSort.js
```diff
@@ -4,9 +4,24 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultCompare) {
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
```
DarkMistyRoom, line 9: Good job, but we can reduce the complexity of the algorithm. You don't need to do the next iteration in the outer loop if there are no changes in the previous one. Consider using do while loop with a counter variable for detecting changes. Feel free to write in the chat, if you have questions

DarkMistyRoom (changes requested):

VitaliyBodnar (approved):

PR #1347, opened 2023-07-11
Added task solution

ruslan-ihnatenko:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 1; i < this.length; i++) {
+      for (let j = 0; j < this.length - 1; j++) {
```
polosanya, line 9: What's the purpose of two nested loops? Actually, one loop is enough

src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
...
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+
+        } else if (this[j].toString() > this[j + 1].toString()) {
+          const temp = this[j];
+
+          this[j] = this[j + 1];
+          this[j + 1] = temp;
```
polosanya, line 22: It would be great to create a separate function (default compare function) and use it as a default value for compareFunction

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/arrayMethodSort.js
ruslan-ihnatenko: Its default buble sort algorithm , without second loop - result will be incorect

loralevitska: you can use [`do while` loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/do...while), if you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/arrayMethodSort.js
```diff
@@ -4,9 +4,45 @@
...
+
+//           this[j] = this[j + 1];
+//           this[j + 1] = temp;
+//         }
+//       }
+//     }
+
+//     return this;
+//   };
+// }
```
loralevitska, line 46: Remove comment. Try it write obvious code that explain all we need without comments)

loralevitska (approved):
Well done) Check my comments for improvement

PR #1348, opened 2023-07-12
Solution arrayMethodSort

Lazarend:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,25 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (item1,
+    item2) => String(item1) > String(item2)) {
```
Viktor-Kost, line 8: move this default value for compareFunction into an own function for better readability

Viktor-Kost (approved):
Nice!
Just apply recommendation from a comment

PR #1350, opened 2023-07-13
js_array-method-sort

KarolinaWiktoria:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
...
+
+    const compareAsString = (a, b) => {
+      if (String(a) > String(b)) {
+        return 1;
+      } else if (String(a) === String(b)) {
+        return 0;
+      } else {
+        return -1;
+      }
+    };
```
DorotaLeniecDev, line 18: You can simplify:
```js
function customCompare(a, b) {
  return a.toString() > b.toString();
}

```
And make use of default function parameter for compareFunction param

DorotaLeniecDev (changes requested):
Good solution but some improvements needed

src/arrayMethodSort.js
KarolinaWiktoria: Thank you for this comment :). => fixed!

darokrk (approved):
Nicely done 👏🏽  approved 🏆 !

PR #1354, opened 2023-07-16
SortMethod

KseniaSkr:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => String(a) > String(b) ? 1 : -1) {
+    let count;
+
+    do {
+      count = 0;
```
etojeDenys, line 12: we can make this var a boolean, it will increase readability

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #1357, opened 2023-07-17
The task is completed

AndruhaMan:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,22 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (item1, item2) => {
```
TarasHoliuk, line 7: default compare function should also return 0 if values are equal to each other
also, let's define default function higher, instead of doing it in arguments

src/arrayMethodSort.js
```diff
@@ -4,8 +4,22 @@
...
+    if (String(item1) > String(item2)) {
+      return 1;
+    }
+
+    return -1;
+  }) {
+    for (let j = 0; j < this.length - 1; j++) {
```
TarasHoliuk, line 14: it's possible more than 2 iterations are needed to complete the sorting
hint: actually, we don't know how many iterations we have to do, but at least one

TarasHoliuk (changes requested):

Viktor-Kost (approved):

PR #1365, opened 2023-07-22
Solution

outofs:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (
+    compareFunction = (a, b) => a.toString() > b.toString()) {
```
OlEzhA0, line 8: would be better to create function for default compare
```
const defaultCompare = (a, b) => a.toString() > b.toString();

[].__proto__.sort2 = function(compareFunction = defaultCompare) {
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        if (compareFunction(this[j], this[j + 1]) > 0) {
+          [this[j], this[j + 1]] = [this[j + 1], this[j]];
+        }
+      }
```
OlEzhA0, line 14: move `this[j + 1]` to constant

also you can start j-iterations from `i + 1`

OlEzhA0 (changes requested):

DanilWeda (approved):

PR #1366, opened 2023-07-22
solution

IhorVerba:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+
+    return this;
+  };
+}
+
+function compare(a, b) {
+  const firstValue = a.toString();
+  const secondValue = b.toString();
+
+  switch (true) {
```
TarasHoliuk, line 27: Avoid using switch(true), it's generally considered bad practice due to its lack of clarity. Use if/else statements for better readability.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compare) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
```
TarasHoliuk, line 9: what if you need more than 2 iterations to sort the array properly?
use another type of loop for this

TarasHoliuk (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #1367, opened 2023-07-24
add task solution of sort method

PavloMykhalov:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,27 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    const compare = compareFunction || function(a, b) {
```
Anastasiia-Svintsova, line 9: Move the default compare function to a separate variable

src/arrayMethodSort.js
```diff
@@ -6,6 +6,27 @@
...
+      if (a > b) {
+        return 1;
+      }
+
+      return 0;
+    };
+
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        if (compare((this[j]).toString(), this[j + 1].toString()) > 0) {
```
Anastasiia-Svintsova, line 23: You're converting elements to strings before comparison. This will lead to incorrect results if the array contains numbers or other non-string elements. You should pass the original elements to the compare function.

Anastasiia-Svintsova (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => a.toString() > b.toString()) {
     // write code here
```
OleziO, line 9: Remove all comments from the code.

OleziO (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #1375, opened 2023-07-27
solutionMethodSort

anastasiia-burkovska:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = function(a, b) {
+    return a.toString() > b.toString();
+  }) {
```
IvanFesenko, line 9: For better readability default function should be like this
```suggestion
  function defaultCompareFunction (a, b) {
    return a.toString() > b.toString();
  }

  [].__proto__.sort2 = function(compareFunction = defaultCompareFunction) {
```

IvanFesenko (approved):
Well done

PR #1376, opened 2023-07-29
custom sort method

AndrzejPec:


choeqq (approved):
Looks little complicated, but works well 🚀

PR #1379, opened 2023-07-31
add task

VladyslavKolisnyk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,22 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
+    String(a) > String(b)) {
```
volodymyr-soltys97, line 8: The default compare function String(a) > String(b) is not accurate, it would return a boolean instead of -1, 0, or 1, which is expected by the sort function. You should adjust this to function(a, b) { return String(a) > String(b) ? 1 : String(a) < String(b) ? -1 : 0; }

src/arrayMethodSort.js
```diff
@@ -4,8 +4,22 @@
...
+    let count;
+
```
volodymyr-soltys97, line 10: The variable 'count' is not descriptively named. Consider renaming it to something more descriptive like 'didSwap' or 'isSorted'.

volodymyr-soltys97 (changes requested):
Goob job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work

PR #1383, opened 2023-08-01
sort method solution

anpiet:


choeqq (approved):
It took me a while to understand your solution, but I think it works well 😄 

Good job ✅

PR #1384, opened 2023-08-01
implement arrayMethodSort task

oleksandrshtonda:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,27 @@
...
+      if (a > b) {
+        return 1;
+      }
+
+      return 0;
+    };
+
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let q = 0; q < this.length - 1 - i; q++) {
+        if (compare(this[q].toString(), this[q + 1].toString()) > 0) {
```
volodymyr-soltys97, line 22: In the compare function, the elements are always converted to strings. It should only convert to string when no compareFunction is provided.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/arrayMethodSort.js
volodymyr-soltys97: Yes, you need to use String in your compareFunction()

volodymyr-soltys97 (changes requested):

PR #1385, opened 2023-08-02
made array method .sort

ivan-baranovskyi:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
...
+          }
+        } else {
+          if (this[j] < this[i]
+          && typeof this[0] === 'string') {
+            minIndex = j;
+          } else if (this[j].toString().localeCompare(this[i].toString()) < 0
+            && typeof this[0] === 'number') {
+            minIndex = j;
+          }
+        }
```
sTorba24, line 26: here you can just call compareFunction(this[j], this[minIndex]), check value and then swap elements or not
```
```

sTorba24 (changes requested):
Good work! Lets simplify it

src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const length = this.length;
```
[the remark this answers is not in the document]
OleziO: Remove this line

OleziO (changes requested):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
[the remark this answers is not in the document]
sTorba24: still not set default value

src/arrayMethodSort.js
```diff
@@ -1,11 +1,33 @@
...
+            = this[j] < this[i]
+              ? [this[i], this[j]] = [this[j], this[i]]
+              : [this[i], this[j]];
+        } else {
+          [this[i], this[j]]
+            = this[j].toString().localeCompare(this[i]) === -1
+              ? [this[j], this[i]]
+              : [this[i], this[j]];
+        }
+      }
```
sTorba24, line 27: all these checks are based on `is compareFunction undefined or not` 
It won't be if you set default value for argument
create callBack function and pass it as a default value for compareFunction argument

sTorba24: then just call compareFunction

sTorba24: 1 condition that you should check - `swap or not`

sTorba24 (changes requested):

src/arrayMethodSort.js
```diff
@@ -1,12 +1,35 @@
...
+
+  function sorting(a, b) {
+    const areCyrillics = /[\u0400-\u04FF]+/;
+
+    if (areCyrillics.test(a) && areCyrillics.test(b)) {
+      return a < b;
+    }
+
+    return new Intl.Collator().compare(a, b);
+  }
```
VitaliyBondarenko1982, line 32:
```suggestion
  function sorting(a, b) {
   return `${a}` > `${b}` ? 1 : -1;
  };
```
no need special checking on cyrillic. 
should pass all tests with such sorting function.

VitaliyBondarenko1982 (approved):
Good job

PR #1390, opened 2023-08-07
Develop

Vitalii-Fediaiev:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const compareFunc = typeof compareFunction === 'function'
+      ? compareFunction
+      : (a, b) => String(a) < String(b) ? -1 : String(a) > String(b) ? 1 : 0;
```
Anastasiia-Svintsova, line 10: Would be great to move this function to a separate constant and assign it as a default value in arguments 
```
function(compareFunction = defaultCompare) { ...

**Review by Anastasiia-Svintsova — CHANGES_REQUESTED** [review]

File: src/arrayMethodSort.js [hunk 2]
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
```
maxim2310, line 7: compareFunction  must to be a separate func, but not argument of sort func

maxim2310 (changes requested):
Nice and creative move)

agniya-i (approved):

PR #1393, opened 2023-08-08
Add task solution

davinnchii:


FrontKid (approved):
Good job!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) > String(b) ? 1 : 0) {
+    let temp = null;
+
+    for (let i = 0; i < this.length; i++) {
+      for (let n = i + 1; n < this.length; n++) {
+        if (compareFunction(this[i], this[n]) > 0) {
+          temp = this[i];
```
33Orange, line 14:
```suggestion
          const temp = this[i];
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) > String(b) ? 1 : 0) {
+    let temp = null;
```
33Orange, line 9: The variable 'temp' is only used in the loop and should be declared there to limit its scope.
```suggestion
```

33Orange (approved):
Good job

PR #1394, opened 2023-08-08
Method sort

Kacper-Leman:
soultion

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -5,7 +5,57 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const array = [...this];
+
+    this.length = 0;
+
+    let compare = compareFunction;
+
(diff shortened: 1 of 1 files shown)


src/arrayMethodSort.js
```diff
@@ -5,7 +5,57 @@
...
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const array = [...this];
+
+    this.length = 0;
+
+    let compare = compareFunction;
+
+    if (!compareFunction) {
```
AleksanderGelo, line 14: It's a good practice to write helper functions, like this, outside the scope of your `applyCustomSort()`,

```
const compareAsFunction = (a, b) => {
// your code
}
```

Then:
` [].__proto__.sort2 = function(compareFunction = compareAsFunction) {
// your code
}`

EwaKobiela (changes requested):
Comments are not fixed. Please review and push changes

PR #1400, opened 2023-08-09
method-sort_solution

olha-rypich:


anastasiia-tilikina (approved):
GJ!

Suggestion for the future:  If you're using a helper such as `compareStrings` here, there's no need to create it right inside your function. It's better to create it outside. Just keep in mind)

PR #1401, opened 2023-08-10
add task solution

ydem1:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,22 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - 1; j++) {
```
denys-danyliuk, line 10: try to improve your solution, now you do many unnecessary iterations

src/arrayMethodSort.js
```diff
@@ -6,6 +6,22 @@
...
+        const resultCompared = arguments.length === 1
+          ? compareFunction(this[j], this[j + 1])
+          : String(this[j]) > (String(this[j + 1]));
```
denys-danyliuk, line 13: create your own compareFunction and pass it as a default param

denys-danyliuk (approved):

PR #1402, opened 2023-08-10
Solved task

AndriiKuz:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) > String(b) ? 1 : 0) {
```
desevoker, line 8:
```suggestion
  [].__proto__.sort2 = function(
    compareFunction = (a, b) => String(a) > String(b)
  ) {
```
I'm not sure that writing a function directly in parameters list is a good idea. The readability of this is pretty bad.
You can make a small nested comparator function and use it as a default. But it's up to you.

Also, `true` is greater than `0`.

desevoker (approved):
Good job! 🚀

src/arrayMethodSort.js
```diff
@@ -4,8 +4,20 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        if (compareFunction(this[j], this[j + 1]) > 0) {
+          const tempVar = this[j];
```
tiserett, line 12: just call it 'temp', no need to add 'var'

tiserett (approved):

PR #1403, opened 2023-08-10
solution

helenmarth:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => a.toString() > b.toString()
```
Viktor-Kost, line 8: The compareFunction does not return a numeric value. It should return a number depending on the comparison, typically -1, 0, or 1. This can cause incorrect sorting order.

Viktor-Kost (approved):
Nice!
Some minor fixes needed

src/arrayMethodSort.js
Viktor-Kost, line 8: The compareFunction does not return a numeric value. It should return a number depending on the comparison, typically -1, 0, or 1. This can cause incorrect sorting order.

Viktor-Kost (changes requested):

PR #1404, opened 2023-08-10
Solution

desevoker:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,34 @@
...
+          swap(this, i, i + 1);
+          swapped = true;
+        }
+      }
+    } while (swapped);
+
+    return this;
+
+    function defaultComparator(a, b) {
+      return a.toString() > b.toString();
```
33Orange, line 26: The defaultComparator function is not correctly implemented. In Javascript, the '>' operator returns a boolean, not a number. The correct implementation should be `return String(a) > String(b) ? 1 : -1;`

33Orange (changes requested):
Almost done

src/arrayMethodSort.js
desevoker: And it doesn't change the sort function behavior at all.

vserhiichuk (approved):
Good job!

src/arrayMethodSort.js
```diff
@@ -1,11 +1,34 @@
...
- * Implement method Sort
- */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (this.length < 2) {
+      return this;
+    }
+
+    const comparator = compareFunction || defaultComparator;
```
BudnikOleksii, line 9: It's better to use [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) instead of assign with "or" operator

src/arrayMethodSort.js
```diff
@@ -1,11 +1,34 @@
...
+
+    return this;
+
+    function defaultComparator(a, b) {
+      return a.toString() > b.toString() ? 1 : -1;
+    }
+
+    function swap(arr, i, j) {
+      [arr[j], arr[i]] = [arr[i], arr[j]];
+    }
```
BudnikOleksii, line 31: Don't create functions inside other functions, move helpers outside.

BudnikOleksii (changes requested):
Good job!
Don't create functions inside other functions, move helpers outside.

polosanya (approved):
👍

PR #1406, opened 2023-08-12
Solution

Orpheus29:


KatOlista (approved):
good job!

src/arrayMethodSort.js
```diff
@@ -4,9 +4,18 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
-  };
+  [].__proto__.sort2
+    = function(compareFunction = (a, b) => a.toString() > b.toString()) {
```
33Orange, line 8: The compareFunction should return a number, not a boolean. You should subtract 'b' from 'a' instead of comparing them with '>'.

33Orange (changes requested):
Almost done

src/arrayMethodSort.js
Orpheus29: Thank you! Is it better to use String() in this task rather than .toString() and why?

33Orange: @Orpheus29 
[if an identifier refers to null or undefined, you can't use the toString method](https://stackoverflow.com/questions/3945202/whats-the-difference-between-stringvalue-vs-value-tostring)

33Orange (approved):
Well done

PR #1407, opened 2023-08-13
add task solution

vserhiichuk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => a.toString() > b.toString()
```
polosanya, line 8: The default compare function should return a number, not a boolean. It should return a negative, zero, or positive value depending on the arguments

BudnikOleksii (approved):
Good job!

polosanya (approved):
Nice!

PR #1408, opened 2023-08-13
add solution

KatOlista:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,20 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
+    a.toString() > b.toString()) {
```
33Orange, line 5: The compareFunction should return a number, not a boolean. You should subtract 'b' from 'a' instead of comparing them with '>'.

src/arrayMethodSort.js
```diff
@@ -1,11 +1,20 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        const elem = this[i];
+
+        if (compareFunction(this[i], this[j]) > 0) {
+          this[i] = this[j];
+          this[j] = elem;
+        }
```
33Orange, line 13:
```suggestion
        if (compareFunction(this[i], this[j]) > 0) {
          [this[i], this[j]] = [this[j], this[i]];
        }
```

33Orange (changes requested):
Almost done

33Orange (changes requested):
.

src/arrayMethodSort.js
```diff
@@ -1,11 +1,17 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
+    a.toString() > b.toString() ? 1 : -1) {
```
Moroz-Dmytro, line 5: Better will be to create a constant value for this default callback and then use it here. It will be easier to read

Moroz-Dmytro (approved):
LGTM 👍 
Small advice

PR #1410, opened 2023-08-14
solution

00Mass00:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,17 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
+    String(a) > String(b)) {
```
Moroz-Dmytro, line 5: The compareFunction should return a `number`, not a `boolean`. You should subtract 'b' from 'a' instead of comparing them with '>'.

src/arrayMethodSort.js
```diff
@@ -1,11 +1,17 @@
...
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        if (compareFunction(this[j], this[j + 1]) > 0) {
+          [this[j], this[j + 1]] = [this[j + 1], this[j]];
+        }
+      }
+    }
```
Moroz-Dmytro, line 12:
```suggestion
    for (let i = 0; i < this.length; i++) {
      for (let j = i + 1; j < this.length; j++) {
        if (compareFunction(this[i], this[j]) > 0) {
          [this[i], this[j]] = [this[j], this[i]];
        }
      }
    }
```

src/arrayMethodSort.js
Moroz-Dmytro: Also would be better to create a separate default compare function an use it here

Moroz-Dmytro (changes requested):
Good work 👍 
Some comments below

roman-mirzoian (approved):
Well done 👍

PR #1412, opened 2023-08-16
Solution

ValeriiaFomenko:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,23 @@
...
-    // write code here
+    const array = this;
+
+    for (let i = 0; i < array.length - 1; i++) {
+      for (let j = 0; j < array.length - i - 1; j++) {
+        const a = array[j];
+        const b = array[j + 1];
+        const compareValue = compareFunction
+          ? compareFunction(a, b)
+          : String(a) > String(b);
```
VitaliyBodnar, line 16: it's better to `String(a) > String(b)` as default value to the compareFunction argument

VitaliyBodnar (changes requested):
Almost done!

etojeDenys (approved):
lgtm

PR #1414, opened 2023-08-16
add sort method solution

Livan94:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const defaultCompare = (a, b) => String(a) > String(b);
```
lerastarynets, line 8: The default compare function may not work as expected as it compares strings, not numbers. It can lead to incorrect results for numeric sorting.

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #1418, opened 2023-08-18
Solution

Nikita211112:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,24 @@
...
+  [].__proto__.sort2 = function(compareFunction
+  = (a, b) => String(a) > String(b)) {
+    const array = this;
+    const n = array.length;
+
+    for (let i = 0; i < n; i++) {
+      for (let j = 0; j < n - i - 1; j++) {
+        const compareResult = compareFunction
+          ? compareFunction(array[j], array[j + 1])
+          : String(array[j]).localeCompare(array[j + 1]);
```
etojeDenys, line 16: why do you need this check if you are adding a default value for compareFunction

etojeDenys (changes requested):

src/arrayMethodSort.js
Nikita211112: > why do you need this check if you are adding a default value for compareFunction
Thx , done

etojeDenys (approved):
lgtm

PR #1426, opened 2023-08-21
Solution

SergiyVS:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+  [].__proto__.sort2 = function(compareFunction = function(a, b) {
+    if (b.toString() === a.toString()) {
+      return 0;
+    };
+
+    return a.toString() > b.toString()
+      ? 1
+      : -1;
+  }) {
+    let x = 0;
```
lerastarynets, line 16: Variable 'x' is not named descriptively. Consider using a more descriptive name, such as 'temp' or 'swap' to indicate its purpose.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit
You can simplify your solution with just to nested `for` loops

src/arrayMethodSort.js
```diff
@@ -3,9 +3,41 @@
...
+      return 0;
+    };
+
+    return a.toString() > b.toString()
+      ? 1
+      : -1;
+  }) {
+    let swap = 0;
+
+    while (!isSorted(this, compareFunction)) {
```
witflash, line 28: Do you really need to use `while` here?
You used `bubble sorting` algorithm, so it's enough to use two nested cycles. There is no need to check after your cycles had done if your list is sorted. It should be definitely sorted already.

witflash (changes requested):
Good job!
But could you check if you really need to use `while` in your code?

TarasHoliuk (approved):

PR #1429, opened 2023-08-22
add task solution

KlymSudakov:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,18 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => a.toString() > b.toString()
```
roman-mirzoian, line 5: The default compare function should return a number, not a boolean. It should return a negative, zero, or positive value depending on the arguments.

roman-mirzoian (approved):
Well done, but pay attention to the comment.

PR #1431, opened 2023-08-23
competed task

TKuts:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,18 @@
  * Implement method Sort
```
OleziO, line 4: Remove all comments from the code.

OleziO (changes requested):

maxim2310 (approved):
Well done!!!

PR #1433, opened 2023-08-24
solution

holovkevych:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
IvanFesenko, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b)) {
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,17 @@
...
-    // write code here
+    const newCompare = compareFunction || ((a, b) => String(a) > String(b));
```
IvanFesenko, line 8:
```suggestion

```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,17 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (newCompare(this[i], this[j]) > 0) {
```
IvanFesenko, line 12:
```suggestion
        if (compareFunction(this[i], this[j]) > 0) {
```

IvanFesenko (approved):
No need to create separate variable, you can set default value if function was not given

PR #1435, opened 2023-08-25
add task solution

Danylo-Chumakov:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const start = this.slice();
```
Anastasiia-Svintsova, line 8: The 'start' variable is not very descriptive. Consider renaming it to something more meaningful, like 'copiedArray' or 'sortedArray'. And BTW this variable seems redundant, you can use this() instead. The sort method is mutable

src/arrayMethodSort.js
```diff
@@ -5,7 +5,24 @@
...
+
+        if (shouldSwap) {
+          const temp = start[j];
+
+          start[j] = start[j + 1];
+          start[j + 1] = temp;
+        }
+      }
+    }
+    this.splice(0, this.length, ...start);
```
Anastasiia-Svintsova, line 23: Don't use array methods in this task.

Anastasiia-Svintsova (changes requested):
Hey! Try to refactor the code according to my comments

volodymyr-soltys97 (approved):
Well done 🔥

PR #1438, opened 2023-08-25
Sort solution

MTS13:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        const swap = compareFunction
+          ? compareFunction(this[j], this[j + 1]) > 0
+          : String(this[j]) > String(this[j + 1]);
```
loralevitska, line 12: Casting array elements to a String can lead to unexpected results when sorting numbers. It's better to handle numbers and strings separately to avoid wrong sorting order.

loralevitska (approved):
Good job!

PR #1440, opened 2023-08-28
added task solution

olya-girnyak:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,21 @@
 /**
  * Implement method Sort
  */
+const compareStrings = (a, b) => {
+  return String(a) > String(b) ? 1 : 0;
```
Viktor-Kost, line 7: The comparison in 'compareStrings' function is incorrect. It should return -1 if 'a' is less than 'b', 0 if they are equal, and 1 if 'a' is greater than 'b'.

Viktor-Kost (changes requested):
Good job!
Just a minor fix needed

Viktor-Kost (approved):

PR #1447, opened 2023-08-31
the task was complete

Viktoria-Moroz:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,36 @@
 /**
  * Implement method Sort
  */
+
+function convertToString(a, b) {
```
loralevitska, line 7: This is not converting this comparing function. Improve function naming

loralevitska (approved):
Good job! Check my comment for improvement

PR #1456, opened 2023-09-04
add task solution

mishakoshilka:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
+        for (let j = 0; j < this.length - i - 1; j++) {
+          if (String(this[j]) > String(this[j + 1])) {
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
+      }
+    }
```
volodymyr-soltys97, line 30: No need to write if else construction here, create a compare function and pass it as params to this task

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's to improve your code

maxim2310 (approved):
Good work!!!

PR #1457, opened 2023-09-05
add task solution

Daniil-A4:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,27 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
VitaliyBodnar, line 7: let's set a default value for compare function as it will make a code shorter 
`[].__proto__.sort2 = function(`
 ` compareFunction = (a, b) => String(a) > String(b) ? 1 : -1`
 `) {`

src/arrayMethodSort.js
```diff
@@ -5,7 +5,27 @@
...
+        } else {
+          if (this[j].toString() > this[j + 1].toString()) {
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
+      }
+    }
```
VitaliyBodnar, line 26:
```suggestion
    for (let i = 0; i < this.length - 1; i++) {
      for (let j = 0; j < this.length - 1 - i; j++) {
        if (compareFunction(this[j], this[j + 1])) {
            const temp = this[j];

            this[j] = this[j + 1];
            this[j + 1] = temp;
         }
      }
    }
```

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):

PR #1459, opened 2023-09-06
Tests done

momos1703:


src/arrayMethodSort.js
```diff
@@ -2,11 +2,40 @@
 
 /**
  * Implement method Sort
- */
+//  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let value1 = '';
+    let value2 = '';
```
Viktor-Kost, line 9: redundant
```suggestion
```

src/arrayMethodSort.js
```diff
@@ -2,11 +2,40 @@
...
+
+    if (!compareFunction) {
```
Viktor-Kost, line 11: to avoid this condition you can add a default parameter for compareFunction to compare as strings
```(a, b) => String(a) > String(b)```

```suggestion
```

Viktor-Kost (changes requested):

src/arrayMethodSort.js
momos1703: Done!

src/arrayMethodSort.js
momos1703: It's very clever. Done!

src/arrayMethodSort.js
```diff
@@ -2,11 +2,24 @@
 
 /**
  * Implement method Sort
- */
+//  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) > String(b)) {
```
anastasiia-tilikina, line 8: move your compare function to a separate method to make the code clearer.

anastasiia-tilikina: and let's make this function return 0, 1, or -1 as the original sort function

src/arrayMethodSort.js
```diff
@@ -2,11 +2,24 @@
...
+    for (let a = this.length; a >= 0; a--) {
+      for (let i = 0; i < this.length - 1; i++) {
+        const value1 = this[i];
+        const value2 = this[i + 1];
+
+        if (compareFunction(value1, value2) > 0) {
+          this[i] = value2;
+          this[i + 1] = value1;
```
anastasiia-tilikina, line 16:
```suggestion
          [this[i + 1], this[i]] = [this[i], this[i + 1]]
```
Here is small tip on how to swap elements using a restructuring assignment

anastasiia-tilikina (changes requested):
almost done!

src/arrayMethodSort.js
momos1703: Done

src/arrayMethodSort.js
momos1703: Frankly saying, I can't understand how to solve it...

src/arrayMethodSort.js
```diff
@@ -2,11 +2,20 @@
 
 /**
  * Implement method Sort
- */
+//  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) >= String(b) ? 1 : -1) {
```
VitaliyBondarenko1982, line 8: answer to your question can be something like that)
defaultCompareFuncion create above applyCustomSort

```
  const strA = String(a);
  const strB = String(b);

  switch (true) {
    case strA > strB:
      return 1;
    case strA < strB:
      return -1;
    default:
      return 0;
  }
};
```
```suggestion
  [].__proto__.sort2 = function(compareFunction = defaultCompareFunction) {
```

VitaliyBondarenko1982 (changes requested):
Please check comment and try to fix prev comment.

src/arrayMethodSort.js
momos1703: Hi) I know, how to create separate method, sorry for this confuse. I can't understand how this method use as default for compareFunction. I can't just add it as parameter, it's not working( 

`[].__proto__.sort2 = function(compareFunction = defaultCompareFunction ) {`

lerastarynets (approved):
All requirements are met.

src/arrayMethodSort.js
momos1703: It's absolutely clear now, maybe even too much) Thank you!

PR #1460, opened 2023-09-06
solution task

Butovych:
I tried to describe all options, but I understand that the solution is far from ideal. Ready for lots of fixes!

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -5,8 +5,83 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let copyCompare = compareFunction;
+
+    if (compareFunction === undefined && typeof this[0] === 'string') {
+      copyCompare = (firstElement, secondElement) =>
+        firstElement > (secondElement);
+
+      do {
+        for (let i = 0; i < this.length - 1; i++) {
+          const first = '' + this[i];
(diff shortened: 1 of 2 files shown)


src/arrayMethodSort.js
```diff
@@ -5,8 +5,83 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let copyCompare = compareFunction;
```
Viktor-Kost, line 8: to avoid such a complicated solution let's make next steps:
1.  add a default parameter for compareFunction to compare as strings
```(a, b) => String(a) > String(b)```

2. use loop inside another loop to work with elements
```
```

3. inside inner loop create a condition using `compareFunction`
```
if (compareFunction(this[j], this[j + 1]) > 0) {
```

4. if we pass a condition switch values
```
const temp = this[j + 1];

this[j + 1] = this[j];
this[j] = temp;
```

5. after loops just return `this`

Viktor-Kost (changes requested):

src/arrayMethodSort.js
Butovych: Thank you, it turned out much better. But I don't understand why the function "compareFunction(this[j], this[j + 1]) > 0" returns a number instead of a boolean value.

mykhalenych (approved):

PR #1462, opened 2023-09-10
task solution

BeginerAlex:


etojeDenys (changes requested):
1. do not forget that we can specify the default value to compare function
2. A comparison function is just a function that compares two elements without loops
example:
```
```

PR #1465, opened 2023-09-12
solution

Liubomyr19:
Guideline](https://github.com/Liubomyr19/js_task-guideline/blob/master/README.md)

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -5,7 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let callback = compareFunction;
+
+    if (typeof callback !== 'function') {
+      callback = (a, b) => {
+        if (a === b) {
+          return 0;
+        }
+
+        return a + '' < b + '' ? -1 : 1;
+      };
+    }
+
+    for (let i = 0; i < this.length; i++) {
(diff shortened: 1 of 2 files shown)


src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
...
+
+    if (typeof callback !== 'function') {
+      callback = (a, b) => {
+        if (a === b) {
+          return 0;
+        }
+
+        return a + '' < b + '' ? -1 : 1;
+      };
+    }
```
VitaliyBodnar, line 18:
```suggestion
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
VitaliyBodnar, line 7: you can make it shorter 
```suggestion
  [].__proto__.sort2 = function(
    compareFunction = (a, b) => (String(a) > String(b) ? 1 : -1)) {
  ) {
```

VitaliyBodnar (changes requested):
Almost done!

src/arrayMethodSort.js
Liubomyr19: Thanks!

VitaliyBodnar (approved):
Good job!

PR #1466, opened 2023-09-12
task solution

no4kar:


src/arrayMethodSort.test.js
```diff
@@ -43,7 +43,7 @@ test(`sort strings: ['c','b','a','d'].sort2()`, () => {
 
 test(`sort cyrillic letters: ['ёж','як','аист'].sort2()`, () => {
   expect(['ёж', 'як', 'аист'].sort2())
-    .toEqual(['аист', 'як', 'ёж']);
+    .toEqual(['аист', 'ёж', 'як']);
```
sTorba24, line 46: why did u change it?

sTorba24 (approved):

src/arrayMethodSort.test.js
no4kar: may be i wrong, but as I remember "я" at the end of the alphabet

no4kar: i thought that is a error, but if need repeat all( and undesirable) behaviors of std sort(), i made correction

PR #1479, opened 2023-09-19
added task solution

koros-rk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    return a.toString() > b.toString() ? 1 : -1;
+  }) {
```
StasSokolov, line 9: let's create a separated function for this logic

src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (compareFunction(this[i], this[j]) > 0) {
+          const tmp = this[i];
+
+          this[i] = this[j];
+          this[j] = tmp;
+        }
```
StasSokolov, line 17: let's use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) for swapping values.

StasSokolov (changes requested):
Good job, let's improve your code a bit more!

StasSokolov (approved):
Awesome🔥

PR #1482, opened 2023-09-19
array method sort

whiovt:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,22 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
-  };
+  // write code here
+  [].__proto__.sort2
+    = function(compareFunction = (a, b) => String(a) > String(b)) {
```
anastasiia-tilikina, line 9: let's create a separate compare function instead of creating it directly in method's parameter in order to keep our code clear.

anastasiia-tilikina: Also now your function returns boolean.
Our original sort method accepts functions that return's either 1, -1 or 0
Let's make it work like this as well

src/arrayMethodSort.js
```diff
@@ -4,9 +4,22 @@
...
+      for (let i = 0; i < this.length; i++) {
+        for (let k = i + 1; k < this.length; k++) {
+          if (compareFunction(this[i], this[k]) > 0) {
+            const compare = this[i];
+
+            this[i] = this[k];
+            this[k] = compare;
+          }
+        }
+      }
```
anastasiia-tilikina, line 19: let's use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) for swapping values.

image

PR #1484, opened 2023-09-21
solution

OlegPopovych:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,37 @@
...
+        return -1;
+      }
+
+      if (String(a) > String(b)) {
+        return 1;
+      }
+
+      if (String(a) === String(b)) {
+        return 0;
+      }
```
StasSokolov, line 19: let's use ternary operator for this `-1 : 1`

StasSokolov (changes requested):
Good job, let's improve your code a bit more!

src/arrayMethodSort.js
OlegPopovych: Thanks, it's done.

StasSokolov (approved):
Awesome job🔥

PR #1491, opened 2023-09-25
Dev

donizer:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+        if (String(a) < String(b)) {
+          return -1;
+        }
+
+        if (String(a) > String(b)) {
+          return 1;
+        }
+
+        return 0;
+      };
```
anastasiia-tilikina, line 25: let's move your default fn outside your method and assign it as the default value for the parameter. In this case you can avoid checking type.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    }
+
+    const length = this.length;
+
+    for (let i = 0; i < length - 1; i++) {
+      for (let j = 0; j < length - 1 - i; j++) {
+        if (compareFn(this[j], this[j + 1]) > 0) {
+          [this[j], this[j + 1]] = [this[j + 1], this[j]];
```
anastasiia-tilikina, line 33: great!

anastasiia-tilikina (changes requested):
well done! let's jus add some fixes

ivbutsykin (approved):
Good job!

PR #1496, opened 2023-09-25
add task solution

Victoriia-Melnyk:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,33 @@
...
+        for (let j = i + 1; j < this.length; j++) {
+          if (this[i].toString() > this[j].toString()) {
+            const temp = this[i];
+
+            this[i] = this[j];
+            this[j] = temp;
+          }
+        }
+      }
+    }
```
etojeDenys, line 19: you can just add some default compare function instead of this logic

etojeDenys (changes requested):
feel free to ask me in chat in you have some questions

src/arrayMethodSort.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) > String(b)) {
+    for (let i = 0; i < this.length; i++) {
```
VitaliyBodnar, line 9:
```suggestion
  [].__proto__.sort2 = function(compareFunction =
    (a, b) => String(a) > String(b)
  ) {
    for (let i = 0; i < this.length; i++) {
```

VitaliyBodnar (changes requested):
code is not aligned

VitaliyBodnar (approved):

PR #1503, opened 2023-09-28
Bubble sort

ElenaGrasovskaya:


mpohorenyi (approved):
Nice work!

src/arrayMethodSort.js
```diff
@@ -1,11 +1,22 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    return (a.toString() >= b.toString()) ? 1 : -1;
```
polosanya, line 5: Don't write it directly here. Let's create a separate default compare function

src/arrayMethodSort.js
```diff
@@ -1,11 +1,22 @@
...
+  }) {
+    if (!this) {
+      return this;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        if (compareFunction(this[j], this[j + 1]) > 0) {
+          [this[j + 1], this[j]] = [this[j], this[j + 1]];
```
polosanya, line 14: 👍

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comment)

src/arrayMethodSort.js
ElenaGrasovskaya: Thank you! Wanted to avoid adding another let variable ... But it does look ugly, so I took it into consideration and implemented :)

PR #1505, opened 2023-09-28
Develop

HokageHEL:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
+        for (let j = 0; j < this.length - i - 1; j++) {
+          if (String(this[j]) > (String(this[j + 1]))) {
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
+      }
+    }
```
Viktor-Kost, line 30: to avoid using this else condition just create a default compareFunction and use as a default parameter 

Example: 
```
const compareAsString = (a, b) => {
  const stringA = String(a);
  const stringB = String(b);

  if (stringA > stringB) {
    return 1;
  }

  if (stringA === stringB) {
    return 0;
  }

  return -1;
};
```

```
```

Viktor-Kost (changes requested):

maxim2310 (approved):
Good work!!!

PR #1507, opened 2023-09-29
Solution

GlibVK:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (typeof compareFunction === 'function') {
```
VitaliyBodnar, line 8: make sense to set default value for `compareFunction ` because to need to avoid code duplication

VitaliyBodnar (changes requested):

VitaliyBodnar (changes requested):
.

IvanFesenko (approved):
Well done

PR #1508, opened 2023-09-29
add task solution

mshmigelska:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,22 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => `${a}` > `${b}`) {
+    for (let i = 0; i < this.length - 1; i++) {
```
anastasiiavorobiova, line 8: There is no need to use an additional for cycle if you don't use ```i```

anastasiiavorobiova (changes requested):
Good work! Almost done!

src/arrayMethodSort.js
Andriy-Shymkiv: not fixed. you can just try to pass the tests without this loop. just leave the one with `k`

Andriy-Shymkiv (changes requested):

volodymyr-soltys97 (approved):

PR #1510, opened 2023-09-30
add task solution

tetyana-shatrova:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
...
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    let cf = compareFunction;
+
+    if (cf === undefined) {
+      cf = (a, b) => a.toString() === b.toString()
+        ? 0 : a.toString() > b.toString() ? 1 : -1;
+    }
+
+    let f = true; let t;
```
etojeDenys, line 15: you should add some default value for compareFunction instead of this

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,24 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) === String(b) ? 0 : String(a) > String(b) ? 1 : -1) {
+    let f = true; let t;
```
VitaliyBodnar, line 9: It's a bad practice to name variables with only one letter

VitaliyBodnar (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,25 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => String(a) === String(b) ? 0 : String(a) > String(b) ? 1 : -1) {
```
IvanFesenko, line 8:
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b) {
```

IvanFesenko (approved):

PR #1520, opened 2023-10-07
Solution

kseniia-chepur:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,22 @@
...
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        const currentEl = this[j];
+
+        const comparisonResult = compareFunction
+          ? compareFunction(this[j], this[j + 1])
+          : `${this[j]}` > `${this[j + 1]}`;
```
VitaliyBodnar, line 14: it's better to set default value for compareFunction in function parameters 
` [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b)) {`

VitaliyBodnar (changes requested):

src/arrayMethodSort.js
kseniia-chepur: Ok, fixed

VitaliyBodnar (approved):
Good job!

PR #1524, opened 2023-10-10
Solution

Alla-Karakaliuk:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
...
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i; j < this.length; j++) {
+        let condition;
+
+        if (compareFunction === undefined) {
+          condition = this[i].toString() > this[j].toString();
+        } else {
+          condition = compareFunction(this[i], this[j]) > 0;
+        }
```
VitaliyBodnar, line 16: it's better to set a default value for the compare function as it will make a code shorter and cleaner 
`[].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b)) {`

VitaliyBodnar (changes requested):

src/arrayMethodSort.js
Alla-Karakaliuk: thanck you

VitaliyBodnar (approved):
Good job!

PR #1530, opened 2023-10-12
write a solution

DamianAdamczewski:


witflash (changes requested):
Good job, Damian!

But let's improve your code a little bit.
You can see you have too many duplicated codes. So you can improve it.

The best way to do it is to default the `compareFunction` with your own function. For example:
```
function sortByDefault(a, b) {
  return String(a) > String(b) ? 1 : -1
}
 
function(compareFunction = sortByDefault) {
  ...
}
```

And you can avoid using the `else` block. With such an approach you can use just code from your `if` block.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,25 @@
...
+  // Default comparison function
+  function sortByDefault(a, b) {
+    return String(a) > String(b) ? 1 : -1;
+  }
+
+  [].__proto__.sort2 = function(compareFunction = sortByDefault) {
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (compareFunction(this[i], this[j]) > 0) {
+          // Swap elements if compareFunction returns a positive value
```
nazarmatsevych, line 16: Next time please remove commented code in your final solution

nazarmatsevych (approved):
Great work 🔥

PR #1532, opened 2023-10-13
Homework: JS Array Method Sort

ViktorBykh:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,25 @@
 /**
  * Implement method Sort
  */
+
+const compareStr = (a, b) => `${a}` > `${b}` ? 1 : -1;
```
sarakhmen, line 7: But what if a == b? In this case they are equal and we don't need to swap anything. It's a usual case to return 0, -1 or 1 from the comparison function

sarakhmen (changes requested):
Let's slightly improve your solution :)

sarakhmen (approved):
LGTM!

PR #1533, opened 2023-10-13
Solution array method sort

garmoni:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => (String(a) > String(b))) {
```
loralevitska, line 8: Move default compare function into variable with meaningful variable naming

loralevitska (approved):

PR #1535, opened 2023-10-14
first solution

Clover247:


StasSokolov (changes requested):
Great job, let's improve your code a bit more!

PR #1539, opened 2023-10-15
Sort

stunnerboyone:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,29 @@
...
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    if (String(a) < String(b)) {
+      return -1;
+    }
+
+    if (String(a) > String(b)) {
+      return 1;
+    }
+
+    return 0;
```
nazarmatsevych, line 16: You can move the default compare function to a separate function and pass its name to the default parameter value

nazarmatsevych (approved):
Good work, left minor comment

PR #1540, opened 2023-10-17
js_array-method-sort

Donovanoff:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,26 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => String(a) > String(b)
```
anastasiia-tilikina, line 8: create default function separately and implement it according original callback (it means that fn should return either 1, 0 of -1

src/arrayMethodSort.js
```diff
@@ -4,8 +4,26 @@
...
+
+    do {
+      swapped = false;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        if (compareFunction(this[i], this[i + 1]) > 0) {
+          const temp = this[i];
+
+          this[i] = this[i + 1];
+          this[i + 1] = temp;
```
anastasiia-tilikina, line 20: it is better use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) for swapping

anastasiia-tilikina (changes requested):
GJ

let's just make some fixes)

src/arrayMethodSort.js
```diff
@@ -4,8 +4,30 @@
...
+    } while (swapped);
+
+    return this;
+  };
+
+  const defaultSort = (a, b) => {
+    // eslint-disable-next-line curly
+    if (String(a) < String(b)) return -1;
+    // eslint-disable-next-line curly
+    if (String(a) > String(b)) return 1;
```
VitaliyBodnar, line 28: it's bad practice to use eslint-disable
```suggestion
    if (String(a) < String(b)) {
     return -1;
    };
    if (String(a) > String(b)) {
     return 1
    };
```

VitaliyBodnar (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,26 @@
...
+      for (let i = 0; i < this.length - 1; i++) {
+        if (compareFunction(this[i], this[i + 1]) > 0) {
+          const temp = this[i];
+
+          this[i] = this[i + 1];
+          this[i + 1] = temp;
+          swapped = true;
+        }
+      }
+    } while (swapped);
```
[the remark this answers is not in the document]
Donovanoff: I don't think so

VitaliyBodnar (approved):

PR #1545, opened 2023-10-19
Solution

kate0427:


src/arrayMethodSort.js
```diff
@@ -5,8 +5,21 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
etojeDenys, line 7: you need to add here the default value for compareFunction

src/arrayMethodSort.js
```diff
@@ -5,8 +5,21 @@
...
-    // write code here
-  };
-}
+    const defaultCompareFunction = (a, b) => String(a) < String(b) ? -1 : 1;
+    const LoopCompareFunction = compareFunction || defaultCompareFunction;
```
etojeDenys, line 9: and these logic will be redundant

etojeDenys (changes requested):

maxim2310 (changes requested):
Please fix previous comments

anastasiiavorobiova (approved):
Great work!

PR #1553, opened 2023-10-25
Solution

VictoriaPukalo:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
etojeDenys, line 7: let's just add the default value for this argument, and avoid redundant logic in the function

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
+        for (let j = 0; j < this.length - 1 - i; j++) {
+          if (String(this[j]) > String(this[j + 1])) {
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
+      }
+    }
```
volodymyr-soltys97, line 30: Now this code is practically duplicated, try to create a helper function at the top level and pass it as a function parameter, it will simplify your code

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

etojeDenys (approved):
good job

PR #1554, opened 2023-10-25
Solution Sort

Markianek:


lerastarynets (approved):
Well done! Short and clear, I like it)

PR #1568, opened 2023-11-01
Added method Sort

IonShive:


src/arrayMethodSort.js
```diff
@@ -1,12 +1,24 @@
...
+"use strict";
 
 /**
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (
+    compareFunction = (a, b) => String(a) > String(b)
```
maxim2310, line 8: compareFunction must return number not boolean, and better to create separate defaultCompareFunction

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

IonShive:
I tried to fix it, and thanks for the good answer.

VitaliyBondarenko1982 (approved):
Well done

PR #1569, opened 2023-11-01
solution

yevtok:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,22 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
-  };
+  [].__proto__.sort2
+  = function(compareFunction
+    = (a, b) => String(a) > String(b) ? 1 : -1) {
```
polosanya, line 9: Don't linebreak the assignment. Also, it would be great to create a default compare function and use it here

polosanya (changes requested):

roman-mirzoian (approved):
Well done 🔥

PR #1570, opened 2023-11-02
made solution for sorting method

MrHunterBoi:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,29 @@
...
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
-  };
+  const SORTED = 1;
+  const NOT_SORTED = -1;
+  const MEDIUM = 0;
+
+  [].__proto__.sort2
+  = function(compareFunction = (
```
polosanya, line 12: Don't linebreak the assignment

src/arrayMethodSort.js
```diff
@@ -4,9 +4,29 @@
...
+      (el1, el2) => el1.toString() >= el2.toString() ? SORTED : NOT_SORTED)
```
polosanya, line 13: Let's create a separate default function and use it here

polosanya (changes requested):
Nice! Let's improve your code a bit)

MrHunterBoi:
Oh, thanks for the advice, changed code c:

polosanya (approved):
Great!

PR #1572, opened 2023-11-06
Sort method solution

ruslanagnatenko:


yefimov77 (approved):

src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
 /**
  * Implement method Sort
  */
+
+const compareStrings = (previous, current) => {
+  const prev = previous.toString();
+  const curr = current.toString();
```
ArtemTeslenko, line 9: do not use contraction when naming

src/arrayMethodSort.js
```diff
@@ -3,9 +3,42 @@
...
+
+    for (let j = this.length; !sorted; j--) {
+      sorted = true;
+
+      for (let i = 1; i < j; i++) {
+        const prev = this[i - 1];
+        const current = this[i];
+
+        if (compareFunction(prev, current) > 0) {
+          sorted = false;
```
ArtemTeslenko, line 34: if on some iteration compare function return <= 0, then you will not return to the loop and stop sorting, maybe change condition of loop interrupting on 26 line to iterate by the whole array

ArtemTeslenko (changes requested):
Good job) Please see comments

src/arrayMethodSort.js
```diff
@@ -3,9 +3,46 @@
...
+  }
+
+  return -1;
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareStrings) {
+    let sorted;
```
artemforsoff, line 24: it is better to always assign a value to a variable
`let sorted = false;`
so that the programmer working with this code in the future will understand what is in the variable from the start

artemforsoff (changes requested):
Good job, fix latest remarks

artemforsoff (approved):

PR #1573, opened 2023-11-06
Solution

zaura333:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,42 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  [].__proto__.sort2 = function(compareFunction = compareStrings) {
     // write code here
+    let counter;
+
+    do {
+      counter = 0;
```
jacekkosiba, line 12: since this variable is not really counting (it's value is either 0 or 1) I would suggest changing the name to something like isSorting, and using boolean values (true / false) instead of 0 / 1

src/arrayMethodSort.js
```diff
@@ -4,9 +4,42 @@
...
+          counter++;
+        }
+      }
+    } while (counter > 0);
+
+    return this;
   };
 }
 
+function compareStrings(a, b) {
```
jacekkosiba, line 30: the definition of this function should probably be placed inside the sort2() method, so that the method does not depend on any external functions

jacekkosiba (changes requested):
Just two minor changes and it's perfect in my opinion :)

jacekkosiba (approved):

jacekkosiba (approved):

src/arrayMethodSort.js
```diff
@@ -4,9 +4,42 @@
...
+        const a = this[i];
+        const b = this[i + 1];
+
+        if (compareFunction(a, b) > 0) {
+          this[i] = b;
+          this[i + 1] = a;
+          isSorting = true;
+        }
+      }
+    } while (isSorting === true);
```
mvjl000, line 24: Keep in mind that you do this as well:
```suggestion
    } while (isSorting);
```

mvjl000 (approved):
Nailed it! 🚀

PR #1575, opened 2023-11-06
add task solution

ksyvvkaa:


mykhailoyurkovskyi (approved):
Cool👍👌

artemforsoff (approved):
good solution 
I liked the default function in param 🔥

PR #1577, opened 2023-11-07
solution

jacekkosiba:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,48 @@
...
+          return 1;
+        }
+
+        if (_b > _a) {
+          return -1;
+        }
+
+        return 0;
+      };
+    }
```
Patrykmclaren14, line 28: You can move these lines of code to the new function; that way, the code will be clearer.

Patrykmclaren14 (approved):
The code is good and works properly. My comment is just a suggestion based on how I see the solution.

src/arrayMethodSort.js
```diff
@@ -5,7 +5,48 @@
...
+
+    // variable for do while loop
+    let sorting;
```
JoyCoffeeAddict, line 31: the only change I would do is to rename `let sorting` to `let isSorted`. Basically, for booleans we use prefixes `is` or `has` and then a verb in past tense. reade more [here](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)

JoyCoffeeAddict (approved):
Great work!

PR #1578, opened 2023-11-07
Solution

Patrykmclaren14:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,43 @@
 /**
  * Implement method Sort
  */
+
+function compareAsString(a, b) {
```
zaura333, line 7: You could consider moving this function into the scope of sort2 function to avoid making sort2 dependent on external functions.

src/arrayMethodSort.js
```diff
@@ -3,9 +3,43 @@
...
+  }
+
+  return 0;
+}
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = compareAsString) {
+    let count;
```
zaura333, line 25: Naming this variable something like isSorting would make it role more clear

zaura333 (approved):
Good job! I'm passing down two comments I got on my (basically identical) solution, and think are worth considering.

src/arrayMethodSort.js
mvjl000: I totally agree here. It would also make sense to keep a boolean value here ;)

src/arrayMethodSort.js
mvjl000: I honestly think it can be the way it is.

mvjl000 (changes requested):
Very good solution 👍 
Just one small thing to change

mvjl000 (approved):
Flawless 🫶🏼

PR #1580, opened 2023-11-08
add task solution

kate-prakofyeva:


bertsu02 (approved):
Seems clean to me, well done

src/arrayMethodSort.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const length = this.length - 1;
+
+    for (let i = 0; i < length; i++) {
+      for (let j = 0; j < length - i; j++) {
+        const isCallback = compareFunction
```
JoyCoffeeAddict, line 12:
```suggestion
        const isCallback = compareFunction !== undefined && compareFunction !== null 
```

This makes it more readable, (because your solution looks like you coould pass a boolean as a compareFunction) you can also check if `typeof compareFunction === "function"`
This way you will guarantee that `compareFunction` is indeed a defined function. You can then move this check into separate variable like 
`const isCompareFunction = compareFunction !== undefined && compareFunction !== null  && typeof compareFunction === "function"`

then rename `isCallback` because the value of this variable means if we need to swap two numbers in an array

JoyCoffeeAddict (changes requested):
Almost perfect. 
But, if you have time, try to read about different sorting algorithms. Here you used a bubble sort. Fun fact, JS `Array.prototype.sort` is implemented with [quick sort](https://www.algorytm.edu.pl/algorytmy-maturalne/quick-sort.html) [funny quick sort video](https://www.youtube.com/watch?v=3San3uKKHgg)

src/arrayMethodSort.js
kate-prakofyeva: done

JoyCoffeeAddict (approved):

PR #1583, opened 2023-11-10
Develop

mixelio:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+          const a = '' + this[i];
+          const b = '' + this[j];
+          const change = this[i];
+
+          if (a > b) {
+            this[i] = this[j];
+            this[j] = change;
+          }
+        }
+      }
```
nazarmatsevych, line 20: Create a separate function and inside compare values as strings and use this function as a default value for compareFunction

nazarmatsevych (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,32 @@
...
+        const a = '' + this[i];
+        const b = '' + this[j];
+        const change = this[i];
+
+        if (a > b) {
+          this[i] = this[j];
+          this[j] = change;
+        }
+      }
+    }
```
nazarmatsevych, line 19:
```
        const aStr = a.toString();
        const bStr = b.toString();

        if (aStr < bStr) {
          return -1;
        }

        if (aStr > bStr) {
          return 1;
        }

        return 0;
      };
```

I meant something like this. And then we can set this defaultCompareFunction as a default value for our function or event easier we can do like below:
`compareFunction = (a, b) => a.toString() > b.toString()`

nazarmatsevych (changes requested):

etojeDenys (approved):
good job

PR #1593, opened 2023-11-16
Develop

FIREDAN99:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
etojeDenys, line 7: you can add here the default parametr
image

compareFunciton = defaultCompareFunction

etojeDenys (changes requested):

PR #1600, opened 2023-11-20
Sort method try 1

VolodymyrLiubyn:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,58 @@
 /**
  * Implement method Sort
  */
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (compareFunction === undefined) {
```
Viktor-Kost, line 9: to avoid using this condition you can set compareFunction as default parameter

src/arrayMethodSort.js
```diff
@@ -3,9 +3,58 @@
 /**
  * Implement method Sort
  */
+
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
Viktor-Kost, line 8: create a default function compareAsStrings outside applyCustomSort

```suggestion
  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
```

src/arrayMethodSort.js
```diff
@@ -3,9 +3,58 @@
...
+        copyArray[count] = copyArray[copyArray.length - 1];
+        copyArray.length -= 1; // make the array shorter
+        count = 0; // reset the counter
+      }
+
+      for (let k = 0; k < sortArray.length; k++) {
+        this[k] = sortArray[k]; // change the input array
+      }
+
+      return this;
```
Viktor-Kost, line 34: this is complicated solution. You need just to compare as strings
Example: 
```
const compareAsStrings = (a, b) => {
  if ({set_condition_here}) {
    return 0;
  };

  if ({set_another_condition_here}) {
    return 1;
  };

  return -1;
};
```

Viktor-Kost (changes requested):

src/arrayMethodSort.js
VolodymyrLiubyn: Understood. Fixing it.

src/arrayMethodSort.js
```diff
@@ -3,10 +3,47 @@
...
+};
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  [].__proto__.sort2 = function(compareFunction = compareAsStrings) {
     // write code here
+
+    let changeCount = this.length;
+    let change = 0;
+    let difArray = [...this];
```
etojeDenys, line 25: this array looks redundant, you can just create one temporary variable (btw, this var also is not required, you can implement this function without it)

etojeDenys (changes requested):

src/arrayMethodSort.js
VolodymyrLiubyn: Ok, removed unnecessary elements

anastasiiavorobiova (approved):
Awesome!

PR #1611, opened 2023-11-27
complete task

boytrend0108:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (compareFunction === undefined) {
+      for (let i = 0; i < this.length; i++) {
+        for (let j = 0; j < this.length - 1 - i; j++) {
+          if (this[j].toString() > this[j + 1].toString()) {
```
etojeDenys, line 11: just add this as the default parameter

`function(compareFunction = ...) {`

src/arrayMethodSort.js
```diff
@@ -5,7 +5,33 @@
...
+            const tmp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = tmp;
+          }
+        }
+      }
+
+      return this;
+    }
```
etojeDenys, line 21: and after that, you can remove this loop

etojeDenys (changes requested):

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1617, opened 2023-11-30
add task solution

Vlad4567:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,28 @@
...
+      for (let o = i + 1; o < this.length; o++) {
+        const mainItem = this[i];
+        const checkItem = this[o];
+        let doChange;
+
+        if (typeof compareFunction === 'function') {
+          doChange = compareFunction(mainItem, checkItem);
+        } else {
+          doChange = `${mainItem}` > `${checkItem}`;
+        }
```
VitaliyBodnar, line 18: It's better to set default value for compare function. It will make  a code shorter

VitaliyBodnar (changes requested):
.

anastasiiavorobiova (approved):
Awesome work!

PR #1618, opened 2023-11-30
Solution

yuriiyepikhov:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,31 @@
...
+
+        if (!compareFunction) {
+          if (this[i].toString() > this[j].toString()) {
+            resultOfCompare = 1;
+          } else {
+            resultOfCompare = -1;
+          }
+        } else {
+          resultOfCompare = compareFunction(this[i], this[j]);
+        }
```
etojeDenys, line 21: to avoid this you can

src/arrayMethodSort.js
```diff
@@ -6,6 +6,31 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
etojeDenys, line 7: you can just specify the default parameter here

etojeDenys (changes requested):

nazarmatsevych (approved):
👍

PR #1629, opened 2023-12-07
add sort method solution

vladyslav-zakharov:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    return a.toString() > b.toString();
+  }) {
```
anastasiia-tilikina, line 9: let's move default fn to a separate helper and make it work as the original one. (it means that it should return 1, -1 or 0)

src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
...
+    for (let i = 0; i < this.length; i++) {
+      for (let k = i + 1; k < this.length; k++) {
+        if (compareFunction(this[i], this[k]) > 0) {
+          const valueI = this[i];
+
+          this[i] = this[k];
+          this[k] = valueI;
```
anastasiia-tilikina, line 16: you can use [destructuring assignment](https://javascript.info/destructuring-assignment#array-destructuring) for swapping values

anastasiia-tilikina (changes requested):
well done! check my comments

src/arrayMethodSort.js
Diliradon: it`s a beautiful idea))

src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
  * Implement method Sort
  */
```
OleziO, line 5: Remove all comments from the code.

OleziO (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #1632, opened 2023-12-09
Solution

AndriyBabiy:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,62 @@
...
+            } else if (firstValue.charCodeAt(k) > secondValue.charCodeAt(k)) {
+              inputArray[j + 1] = firstValue;
+              inputArray[j] = secondValue;
+              break;
+            } else {
+              break;
+            }
+          }
+        }
+      }
```
nazarmatsevych, line 32: Looks too complicated and a lot of duplication. You can create a separate function same as in sort method [documentation ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)and set it as a default value for compareFunction

function compareFn(a, b) {
  if (a is less than b by some ordering criterion) {
    return -1;
  } else if (a is greater than b by the ordering criterion) {
    return 1;
  }
  // a must be equal to b
  return 0;
}

nazarmatsevych (changes requested):
If you have any questions - feel free to ask them in fe_chat

src/arrayMethodSort.js
AndriyBabiy: Separate function for comparison created

volodymyr-soltys97 (changes requested):
The code now looks a bit complex and confusing, think about how it can be simplified

src/arrayMethodSort.js
```diff
@@ -5,8 +5,38 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
nazarmatsevych, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = stringCompareElements) {
```

And you can remove this if block
image

nazarmatsevych (changes requested):
Almost done, please check the comment below :)

etojeDenys (approved):
good job

PR #1640, opened 2023-12-15
Solution

SerjDSA:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(cb) {
```
```suggestion
  [].__proto__.sort2 = function(cb = ...) {
```
you can just add the default parameter instead if this if logic

etojeDenys (changes requested):

OleziO (approved):

PR #1645, opened 2023-12-18
add solution

nabokok:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,18 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => (a + '') > (b + '')) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.slice(0, i).length; j++) {
```
sasholk, line 10: interesting solution with slice 🔥

sasholk (approved):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,18 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction =
+  (a, b) => (a + '') > (b + '')) {
```
YegorKorenkov, line 8: please move default compareFunction value to separate variable for better code readability

YegorKorenkov (changes requested):
Good job! Please check one minor remark

src/arrayMethodSort.js
```diff
@@ -4,8 +4,19 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const defaultCompareFunc = (a, b) => (a + '') > (b + '');
+
+  [].__proto__.sort2 = function(compareFunction = defaultCompareFunc) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.slice(0, i).length; j++) {
```
SerhiiKirik, line 11: we can make it better with `do / while` function

src/arrayMethodSort.js
```diff
@@ -4,8 +4,19 @@
...
+        if (compareFunction(this[j], this[i]) === true
```
SerhiiKirik, line 12: what return `compareFunction`?
Boolean or Number?

src/arrayMethodSort.js
```diff
@@ -4,8 +4,19 @@
...
+        || compareFunction(this[j], this[i]) > 0) {
+          [this[i], this[j]] = [this[j], this[i]];
```
SerhiiKirik, line 14: let's split it with variables

SerhiiKirik (changes requested):
Well done 🔥 

src/arrayMethodSort.js
nabokok: I think we need a "while" loop when we don't know the count of iterations but in this case, the "for" loop is better.

src/arrayMethodSort.js
nabokok: @SerhiiKirik I'm not sure I understand you, can you explain?

YegorKorenkov: I think that Sergey meant that these values need to be put into variables to make the code more readable. But you shortened the condition, and now in my opinion everything looks normal
image

YegorKorenkov (approved):
Good job! Approve

PR #1650, opened 2023-12-19
task done

Manankin:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,32 @@
...
+
+      counter++;
+    }
+
+    return this;
   };
 }
 
+function defaultCompare(a, b) {
+  return a.toString() > b.toString();
```
ssaviak, line 32: You also could use "locale compare" to avoid unexpected behavior in other languages.

ssaviak (approved):
Nice!

YegorKorenkov (approved):
Nice work!

PR #1653, opened 2023-12-20
Solution

alexrapin:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,22 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
-  };
+  [].__proto__.sort2
+    = function(compareFunction = (a, b) => a.toString() > b.toString()) {
+      chooseMethod(this, compareFunction);
+
+      return this;
+
+      function chooseMethod(array, method) {
```
Viktor-Kost, line 13: 1. move default parameter outside applyCustomSort
2. no needs to declare chooseMethod function. You can just use its solution

Viktor-Kost (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1654, opened 2023-12-25
Solution

p1zzat1me:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
  * Implement method Sort
  */
 function applyCustomSort() {
+
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    for (let i = 0; i < this.length; i++) {
```
l4st1m0za, line 10: The implemented sorting algorithm doesn't seem to be correct. It only makes one pass over the array, hence won't sort the array if it requires more than one pass. Consider implementing a correct sorting algorithm like Bubble Sort, Selection Sort, etc.

src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
...
+      if (compareFunction) {
+        const comparisonRes = compareFunction(this[i], this[i + 1]);
+        if (comparisonRes > 0) {
+          const temp = this[i];
+          this[i] = this[i + 1];
+          this[i + 1] = temp;
+        }
+      }
+      else {
```
l4st1m0za, line 19: The else statement does not have a corresponding if statement. Also, the logic in the else block is incorrect for sorting when no compareFunction is provided. It should sort based on string conversions of the elements, not just using localeCompare.

src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
...
+          this[i] = b;
+          this[i + 1] = a;
+        }
+      }
+      }
+    }
   };
 }
 
+
```
l4st1m0za, line 33: Multiple empty lines at the end of the file are unnecessary and could be removed for cleaner code.

l4st1m0za (changes requested):
Great job, see the comments below

PR #1656, opened 2023-12-26
Develop

ivankovbohdan:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (prev, curr) =>
+    (String(prev) > String(curr))) {
```
nazarmatsevych, line 8: It's better to create a separate function and use it here as a default value for a compareFunction

nazarmatsevych (approved):
Great work, left a minor comment

PR #1661, opened 2023-12-30
solution

vSainiuk:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,21 @@
...
+        }
+      }
+    }
+
+    return this;
   };
 }
 
+function defaultComparison(a, b) {
+  return a.toString() > b.toString();
```
SerhiiKirik, line 21: use `localCompare` for it
Also, `callback` of `sort` returns a number
Check it [here](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)

src/arrayMethodSort.js
```diff
@@ -4,9 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultComparison) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < i; j++) {
+        if (compareFunction(this[j], this[i]) > 0) {
+          [this[i], this[j]] = [this[j], this[i]];
```
SerhiiKirik, line 11: split it with a variables for more readable
```suggestion
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,21 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultComparison) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < i; j++) {
```
SerhiiKirik, line 9: better to do it with `do / while` loop

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your code 🧑‍💻

PR #1663, opened 2024-01-01
Solution of array sort method

Voloshynd:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,23 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => `${a}` > `${b}` ? 1 : -1
```
denys-danyliuk, line 8: would be better to move this default value to separate function

src/arrayMethodSort.js
```diff
@@ -4,8 +4,23 @@
...
+  ) {
+    let stateCondition;
+
+    do {
+      stateCondition = false;
+
+      for (let i = 0; i < this.length - 1; i++) {
+        if (compareFunction(this[i], this[i + 1]) > 0) {
```
denys-danyliuk, line 16:
```suggestion
        if (compareFunction(this[i], this[i + 1])) {
```

denys-danyliuk (approved):

PR #1674, opened 2024-01-10
Solution

SashaViun:


src/arrayMethodSort.js
```diff
@@ -6,6 +6,26 @@
...
     // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        let compareResult;
+
+        if (compareFunction) {
+          compareResult = compareFunction(this[j], this[j + 1]) > 0;
+        } else {
+          compareResult = String(this[j]) > String(this[j + 1]);
+        }
```
IvanFesenko, line 17: Add default value for `compareFunction` instead of if/else

IvanFesenko (changes requested):
Good job. Lets make improvement - check my comment

PR #1675, opened 2024-01-11
Solution

DenLys21:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
```
anastasiia-tilikina, line 7: move default callback to a separate function and implement it the same way as original.
original function should return 1, 0, or -1

read about sort method [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description)

src/arrayMethodSort.js
```diff
@@ -4,8 +4,20 @@
...
+    String(a) > String(b)) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        if (compareFunction(this[i], this[j]) > 0) {
+          const element = this[i];
+
+          this[i] = this[j];
+          this[j] = element;
```
anastasiia-tilikina, line 15: let's use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description) to swap values

anastasiia-tilikina (commented):
GJ

check my comments

anastasiia-tilikina (changes requested):
.

src/arrayMethodSort.js
```diff
@@ -5,7 +5,23 @@
...
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i + 1; j < this.length; j++) {
+        let compareResult;
+
+        if (compareFunction) {
+          compareResult = compareFunction(this[i], this[j]) > 0;
+        } else {
+          compareResult = String(this[i]) > String(this[j]);
```
nazarmatsevych, line 15: You can create a separate variable for a default function and set it as a default value for compareFunction

nazarmatsevych (approved):
Good work, left a minor comment

PR #1677, opened 2024-01-12
sort

marichkamt:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
...
+          return -1;
+        }
+
+        if (a > b) {
+          return 1;
+        }
+
+        return 0;
+      };
+    }
```
etojeDenys, line 22: you can just add the default value for the function

src/arrayMethodSort.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
etojeDenys, line 7: here compareFunction =

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,17 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
+    String(a) > String(b)) {
```
loralevitska, line 8: Better move this function to variable and use it only name

const defaultCompareFunction = (a, b) => String(a) > String(b);

```suggestion
  [].__proto__.sort2 = function(compareFunction = defaultCompareFunction) {
```

loralevitska (approved):
Good job!

PR #1680, opened 2024-01-15
Apply Custom Sort function | JS

julia-dronova:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,43 @@
...
+  switch (true) {
+    case prevStr > currStr:
+      return 1;
+
+    case prevStr === currStr:
+      return 0;
+
+    case prevStr < currStr:
+      return -1;
+  }
```
nazarmatsevych, line 43:
```suggestion
  if (prevStr > currStr:) {
    return 1;
  }
  
  if (prevStr < currStr:) {
    return -1;
  }
  
  return 0;
```

we can use if statement in this case

nazarmatsevych (approved):
Well done 👍

PR #1682, opened 2024-01-16
Solution

sergii-stognii:
Array Iteration Methods Implementation: Method Sort

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -5,7 +5,25 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (compareFunction) {
+      for (let i = 0; i < this.length - 1; i++) {
+        for (let j = 0; j < this.length - i - 1; j++) {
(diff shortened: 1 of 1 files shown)


src/arrayMethodSort.js
```diff
@@ -5,7 +5,25 @@
...
+        }
+      }
+    } else {
+      for (let i = 0; i < this.length - 1; i++) {
+        for (let j = 0; j < this.length - i - 1; j++) {
+          if (String(this[j]) > String(this[j + 1])) {
+            [this[j], this[j + 1]] = [this[j + 1], this[j]];
+          }
+        }
+      }
```
etojeDenys, line 23: instead of if else logic, just add the default value for compareFunction
image

etojeDenys (changes requested):
.

src/arrayMethodSort.js
sergii-stognii: I see, will try to.

volodymyr-soltys97 (approved):
Great work 🔥

PR #1683, opened 2024-01-16
Sort

AnnaPshenychnykh:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
...
+      for (let j = 0; j < this.length - i - 1; j++) {
+        if (compareFunction && compareFunction(this[j], this[j + 1]) < 1) {
+          continue;
+        }
+
+        if (compareFunction === undefined
+            && this[j].toString() <= this[j + 1].toString()
+        ) {
+          continue;
+        }
```
etojeDenys, line 18: instead of this strange logic just add the default value for compareFunction

image

`function(compareFunction = ...) {...}`

etojeDenys (changes requested):

PR #1693, opened 2024-01-24
Dima gres feature

Greshch:
Please check it.

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -16,7 +16,6 @@ function applyCustomSort() {
 
     for (let i = 0; i < this.length - 1; i++) {
       let changes = false;
-
       for (let j = 1; j < this.length - i; j++) {
         const prev = this[j - 1];
         const cur = this[j];
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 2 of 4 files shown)


src/arrayMethodSort.js
```diff
@@ -6,6 +6,52 @@
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
     // write code here
+    const isNumber = typeof this[0] === 'number';
+
+    if (compareFunction === undefined) {
+      this.forEach((element, index) => {
+        this[index] = element.toString();
+      });
```
anastasiiavorobiova, line 14: The compare function shouldn't change an array

src/arrayMethodSort.js
```diff
@@ -6,6 +6,52 @@
...
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      let changes = false;
+
+      for (let j = 1; j < this.length - i; j++) {
```
anastasiiavorobiova, line 20:
```suggestion
 for (let i = 0; i < this.length; i++) {
      for (let j = i + 1; j < this.length; j++) {
        if (compareFunction(this[i], this[j]) ...
```

anastasiiavorobiova (changes requested):
Great work! Let's make it perfect!

src/arrayMethodSort.js
Greshch: I think if I have no compare function as an argumet  I should compare my data like strings and then in the end cast my data to numbers.

src/arrayMethodSort.js
Greshch: Ok. I'll fix it.

PR #1694, opened 2024-01-25
sort method implemented

ivanlavrinenko:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        const shouldSwap = typeof compareFunction === 'function'
+          ? compareFunction(this[j], this[j + 1]) > 0
+          : `${this[j]}` > `${this[j + 1]}`;
```
etojeDenys, line 12: instead of this strange logic you can add the default value

src/arrayMethodSort.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
etojeDenys, line 7: here

etojeDenys (changes requested):

OleziO (approved):

PR #1697, opened 2024-01-26
Solution

ayasinska:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,23 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => String(a) > String(b)
```
vlasiuksasha, line 8: The compare function does not correctly handle all cases. It should return a negative, zero, or positive value, not a boolean. A more correct default comparison could be (a, b) => String(a).localeCompare(String(b)).

src/arrayMethodSort.js
```diff
@@ -4,8 +4,23 @@
...
+  ) {
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        const compareResult = compareFunction(this[j], this[j + 1]);
+
+        if (compareResult > 0) {
```
vlasiuksasha, line 14: The comparison inside the 'if' statement is not correct. The compareFunction should return a negative, zero, or positive value, dependent on the arguments. You should check if compareResult is more than 0.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,23 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
```
vlasiuksasha, line 7: Avoid extending the Array.prototype. It's considered bad practice because it can cause unexpected side effects with other code or libraries that may also be trying to extend the Array.prototype with the same method name.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,23 @@
...
+        if (compareResult > 0) {
+          const temp = this[j];
+
+          this[j] = this[j + 1];
+          this[j + 1] = temp;
+        }
+      }
+    }
+
+    return this;
```
vlasiuksasha, line 23: The method modifies the original array. It's better to return a new sorted array, leaving the original array untouched.

vlasiuksasha (changes requested):
check the comments

src/arrayMethodSort.js
ayasinska: https://mate.academy/chats/27190?selectedMessageId=1401459

src/arrayMethodSort.js
ayasinska: https://mate.academy/chats/27190?selectedMessageId=1401459

src/arrayMethodSort.js
ayasinska: https://mate.academy/chats/27190?selectedMessageId=1401459

src/arrayMethodSort.js
ayasinska: https://mate.academy/chats/27190?selectedMessageId=1401459

Viktoriia-V-V (approved):
Not bad

PR #1702, opened 2024-01-28
task solution

Poviakalo:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,24 @@
...
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        const current = this[j];
+        const next = this[j + 1];
+
+        if (!compareFunction && current.toString() > next.toString()) {
+          this[j] = next;
+          this[j + 1] = current;
```
etojeDenys, line 15: instead of this logic, you can add the default value to compareFunction parametr

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #1705, opened 2024-01-31
solution

AlinaOvod:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,19 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(
+    compareFunction = (a, b) => String(a) > String(b)
+      ? 1
+      : (String(a) < String(b) ? -1 : 0)) {
```
etojeDenys, line 10:
```suggestion
    compareFunction = (a, b) => String(a) > String(b)) {
```
we can simplify it

etojeDenys (changes requested):
small comment

PR #1706, opened 2024-02-01
Solution

bla2001:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) =>
+    String(a) > String(b)) {
```
Viktoriia-V-V, line 8: let`s user variables for strings not to change the initial value

Viktoriia-V-V (changes requested):
think about how you can improve the cycles and the conditions in them

4ksandrey (approved):
Accepted, well done

PR #1707, opened 2024-02-01
task done;

Reaffith:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,28 @@
...
+
+    for (let i = 0; i < arrayLength - 1; i++) {
+      for (let j = 0; j < arrayLength - i - 1; j++) {
+        let shouldSwap = false;
+
+        if (compareFunction) {
+          shouldSwap = compareFunction(this[j], this[j + 1]) > 0;
+        } else {
+          shouldSwap = String(this[j]) > String(this[j + 1]);
+        }
```
etojeDenys, line 18: instead of this strange logic just add the default value for compareFunction

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #1709, opened 2024-02-02
add task solution

KaterinaZhlobinskaya:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,39 @@
...
+        if (String(a) < String(b)) {
+          return -1;
+        }
+
+        if (String(a) > String(b)) {
+          return 1;
+        }
+
+        return 0;
+      };
```
etojeDenys, line 20: instead of this strange logic just add the default value to the compareFunction

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #1715, opened 2024-02-05
solution

xeilen:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
Viktor-Kost, line 7: move default compareFunction to a variable
```suggestion
  [].__proto__.sort2 = function(compareFunction = compareByDefault) {
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,29 @@
...
-    // write code here
+    function swapItems(index) {
+      const temp = this[index];
+
+      this[index] = this[index - 1];
+      this[index - 1] = temp;
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (compareFunction) {
```
Viktor-Kost, line 16: so to avoid this comparing create a `compareByDefault ` function that will be used in case we haven't passed an argument

Viktor-Kost (changes requested):

etojeDenys (approved):
lgtm

PR #1719, opened 2024-02-07
add task solution

DenLysiak:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+
+        if (strA < strB) {
+          return -1;
+        } else if (strA > strB) {
+          return 1;
+        }
+
+        return 0;
+      };
+    }
```
etojeDenys, line 23: instead of this logic, just add the default value for compareFunction parameter

src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+      for (let n = 0; n < arrLength; n++) {
+        if (callCompare(this[n], this[n + 1]) > 0) {
+          const temp = this[n];
+
+          this[n] = this[n + 1];
+          this[n + 1] = temp;
+        }
+      }
+      arrLength--;
+    }
```
etojeDenys, line 37: looks like we can simplify it
```suggestion


    for (let i = 0; i < this.length; i++) {
      for (let n = 0; n < this.length - 1 - i; n++) {
        if (callCompare(this[n], this[n + 1]) > 0) {
          const temp = this[n];

          this[n] = this[n + 1];
          this[n + 1] = temp;
        }
      }

    }
```

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1720, opened 2024-02-08
Sort Method

Denys209818:


lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/arrayMethodSort.js
```diff
@@ -5,8 +5,93 @@
...
+
+            arr[i + 1] = temp;
+          }
+        }
+      }
+
+      return arr;
+    }
+
+    //  !!!selection sort for simple items!!!
```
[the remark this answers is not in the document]
Denys209818: It isn't a comment. It's a working algorithm that could replace QuickSort. I can just comment QuickSort and use commented code to launch the algorithm and it will certainly work. I thought that I can try to solve this task in several approaches and they were to work. In general, they work but second or first algorithm is excessive

src/arrayMethodSort.js
```diff
@@ -5,8 +5,93 @@
...
+    // }
+
+    // !!!QuickSort!!!
+    QuickSort(arr, 0, arr.length - 1, compareFunction);
+
+    return arr;
   };
 }
 
+function QuickSort(array, minIndex, maxIndex, compareFunction) {
```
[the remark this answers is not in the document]
Denys209818: It's been already fixed

.github/workflows/test.yml
GUSILLUS, line 1: You don't need to delete files from .github need only work on solution file. Please return it

src/arrayMethodSort.js
```diff
@@ -5,8 +5,93 @@
...
+    //       minIndex = j;
+    //     }
+    //   }
+
+    //   const temp = arr[i];
+
+    //   arr[i] = arr[minIndex];
+
+    //   arr[minIndex] = temp;
+    // }
```
GUSILLUS, line 49: Please send a file on review without unused or commented code. It will be a good practice

GUSILLUS (changes requested):

.github/workflows/test.yml
Denys209818: If I try to push data with those files I get an error about permission access

src/arrayMethodSort.js
```diff
@@ -5,8 +5,70 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const arr = this;
+
+    // !!!bubble sort for dividing option!!!
+    if (compareFunction && arr.length > 1 && compareFunction(arr[0],
```
etojeDenys, line 11:
```suggestion
    if (arr.length > 1)
```
instead of that logic you just add the default value for compareFunction

etojeDenys (approved):
awesome

PR #1726, opened 2024-02-09
Array sort method task solution

SvNadorozhnyi:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,19 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => `${a}` > `${b}`) {
```
loralevitska, line 7: Move default function to separate variable with meaningful variable naming

loralevitska (approved):
Good job!

PR #1728, opened 2024-02-10
add array method sort solution

lbihun:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
...
+          this[j] = temp;
+        }
+
+        if (compareFunction !== undefined
+          && compareFunction(this[j], this[i]) < 0) {
+          const temp = this[i];
+
+          this[i] = this[j];
+          this[j] = temp;
+        }
```
sTorba24, line 23: so you can swap elements twice?
You should use `this[j].toString() < this[i].toString()` as a default param for `compareFunction` to have smth like
```suggestion
        if (compareFunction(this[j], this[i]) < 0) {
          const temp = this[i];

          this[i] = this[j];
          this[j] = temp;
        }
```

sTorba24 (changes requested):
Almost good! Lets improve it

etojeDenys (approved):
lgtm

PR #1729, opened 2024-02-13
Solution

wasik07:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,28 @@
...
+          if (compareFunction(elementFirst, elementSecond) >= 1) {
+            this[j] = elementSecond;
+            this[j + 1] = elementFirst;
+          }
+        } else {
+          if (`${elementFirst}` < `${elementSecond}`) {
+            continue;
+          } else {
+            this[j] = elementSecond;
+            this[j + 1] = elementFirst;
```
etojeDenys, line 23: instead of these conditions just add the default value for compareFunction argument

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1732, opened 2024-02-13
added solution array method sort

nk8bnj:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (compareFunction) {
+      for (let i = 0; i < this.length; i++) {
+        for (let j = i + 1; j < this.length; j++) {
+          if (compareFunction(this[i], this[j]) > 0) {
+            const temp = this[i];
+
+            this[i] = this[j];
+            this[j] = temp;
```
roman-mirzoian, line 15:
```suggestion
            [this[i], this[j]] = [this[j], this[i]];
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
+        }
+      }
+    } else {
+      for (let i = 0; i < this.length; i++) {
+        for (let j = i + 1; j < this.length; j++) {
+          if (this[i].toString() > this[j].toString()) {
+            const temp = this[i];
+
+            this[i] = this[j];
+            this[j] = temp;
```
roman-mirzoian, line 26:
```suggestion
           [this[i], this[j]] = [this[j], this[i]];
```

src/arrayMethodSort.js
```diff
@@ -5,7 +5,31 @@
...
+          }
+        }
+      }
+    } else {
+      for (let i = 0; i < this.length; i++) {
+        for (let j = i + 1; j < this.length; j++) {
+          if (this[i].toString() > this[j].toString()) {
```
roman-mirzoian, line 22: Export this condition as a separate function and use it as the default value of `compareFunction`, then there will be no need to add an if condition and duplicate similar code

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

BudnikOleksii (approved):
Good work 👍

PR #1741, opened 2024-02-16
Solution

kovaltar:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,47 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (this.length < 2) {
+      return this;
+    }
+
+    const isFunc = compareFunction !== undefined ? 1 : 0;
```
Andriy-Shymkiv, line 12: I don't think that you need this variable

src/arrayMethodSort.js
```diff
@@ -5,7 +5,47 @@
...
+      } else {
+        if (compareFunction(startItem, this[i]) > 0) {
+          lessThenStart.push(this[i]);
+        } else if (compareFunction(startItem, this[i]) < 0) {
+          biggerThenStart.push(this[i]);
+        } else {
+          equalStart.push(this[i]);
+        }
+      }
+    }
```
Andriy-Shymkiv, line 38: to be honest, it doesn't have any sense to create two arrays and duplicate logic inside if/else block.
you can make a simple bubble sorting instead. check [this](https://www.geeksforgeeks.org/bubble-sort-algorithms-by-using-javascript/)

Andriy-Shymkiv (changes requested):

src/arrayMethodSort.js
kovaltar: Thanks a lot. I rewrote the function using bubble sort

src/arrayMethodSort.js
kovaltar: Yes, I agree, it can be done without this variable:)

src/arrayMethodSort.js
```diff
@@ -5,7 +5,25 @@
...
+    if (this.length < 2) {
+      return this;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < (this.length - i - 1); j++) {
+        if ((compareFunction !== undefined
+          && (compareFunction(this[j], this[j + 1]) > 0))
+          || (compareFunction === undefined
+          && (this[j].toString() > this[j + 1].toString()))) {
```
GUSILLUS, line 17: Consider thinking about how to improve the readability of these conditionals and move them to separate variables.

GUSILLUS (approved):
Great job! Think about code style improvements for future)

src/arrayMethodSort.js
kovaltar: Done

PR #1746, opened 2024-02-19
Add solution

ViktoriiaRepo:


olduma (approved):
Looks great to me!

src/arrayMethodSort.js
```diff
@@ -5,7 +5,33 @@
...
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
+      }
+
+      return this;
+    } else {
```
EdPirro, line 21: `else` is not needed when returning inside the previous if statement (early return)

src/arrayMethodSort.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (compareFunction && typeof compareFunction === 'function') {
```
EdPirro, line 8: Consider defining a default value for the compare function instead of repeating the loop:

```js
if (!(compareFunction && typeof compareFunction === 'function')) {
  compareFunction = //default compare function; in this case a string comparison
}
for (...) // implement the sort only once
```

This will avoid implementing the same logic twice changing only the condition, and result in a clearer and easier to mantain code

EdPirro (approved):
Great job, @ViktoriiaRepo! You're implementation is correct!

I added just a few details for you to keep in mind for the next implementations, but you did well, keep it up!

PR #1748, opened 2024-02-20
Solution

dkolinska:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,37 @@
...
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
-  };
+  [].__proto__.sort2
+  = function(customCompareFunction = (a, b) => `${a}` > `${b}`) {
+      let compareFunction = customCompareFunction;
+
+      if (typeof compareFunction !== 'function') {
```
l4st1m0za, line 11: The check for the type of compareFunction is unnecessary. The default value of the compareFunction ensures that it will always be a function.

src/arrayMethodSort.js
```diff
@@ -4,9 +4,37 @@
...
+        compareFunction = (a, b) => {
+          if (a < b) {
```
l4st1m0za, line 13: The lines 13-21 are redundant as the default comparison function provided on line 8 already covers this logic.

l4st1m0za (changes requested):
Good job, but some improvement needed!

l4st1m0za (approved):
Good job!

PR #1752, opened 2024-02-21
solve the task

ogerenko:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,36 @@
...
+          return -1;
+        }
+
+        if (String(a) > String(b)) {
+          return 1;
+        }
+
+        return 0;
+      };
+    }
```
etojeDenys, line 22: just add the default value for compareFunction parameter instead of this

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1754, opened 2024-02-21
done

anastasia1383:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+            swapped = true;
+          }
+        } else {
+          if (a > b) {
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+            swapped = true;
+          }
```
BudnikOleksii, line 30: You do the same in both if and else blocks.
Create a default compare function and set it by default if the `compareFunction` isn't passed to the sort

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution a bit

stassokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #1757, opened 2024-02-22
Develop

DenNice-cloud:


AntonSakh (approved):
Nice!

src/arrayMethodSort.js
```diff
@@ -5,7 +5,37 @@
...
+          return 1;
+        }
+
+        if (valueA < valueB) {
+          return -1;
+        }
+
+        return 0;
+      };
+    }
```
BudnikOleksii, line 25: - Move the default compare function outside the sort method and you can just assign it as a default parameter
- Don't use concatenation to create strings, use template strings, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

PR #1759, opened 2024-02-24
Solution

Ivanyn732:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
...
-    // write code here
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        let compareResult;
+
+        if (compareFunction !== undefined) {
+          compareResult = compareFunction(this[j], this[j + 1]);
+        } else {
+          compareResult = (String(this[j]) > String(this[j + 1]));
+        }
```
VitaliyBondarenko1982, line 16: set for defautlt value for  compareFunciton. it compareFunction will not be passed to sort method - will work default value. 
```
```
...
```suggestion
          compareResult = compareFunction(this[j], this[j + 1]);
```

VitaliyBondarenko1982 (changes requested):
Looks good. Suggest some improvement.

src/arrayMethodSort.js
VitaliyBondarenko1982: with default argument you don't need anymore to chekc on undefined compare function. compareFunciton never will be undefined.
```
const compareResult = compareFunction(this[j], this[j + 1]);
```

VitaliyBondarenko1982 (approved):
Approve this time,  but check comment.

PR #1762, opened 2024-02-25
implementation of array method sort

gboreyko:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,28 @@
...
- * Implement method Sort
- */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = (a, b) => {
+    return (String(a) > String(b))
+      ? 1
+      : -1;
+  }) {
```
vadiimvooo, line 8:
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b)) {
```

vadiimvooo (approved):
Well done. You can also simplify the method for little.

src/arrayMethodSort.js
gboreyko: Thank you for you advice

PR #1764, opened 2024-02-26
add task solution

vladarskyi:


src/arrayMethodSort.js
```diff
@@ -5,8 +5,63 @@
...
+          if (compareFunction(previousValue, currentValue) > 0) {
+            this[i] = previousValue;
+            this[i - 1] = currentValue;
+            isSorted = false;
+          }
+        }
+      }
+
+      return this;
+    }
```
BudnikOleksii, line 29: DRY, you do the same operations in both cases, the only difference is a compare function. Create a default compare function and assign it as a default parameter if a function isn't passed to sort

src/arrayMethodSort.js
```diff
@@ -5,8 +5,63 @@
...
+    let previousValue;
+    let isSorted;
+
+    if (compareFunction !== undefined) {
+      while (!isSorted) {
+        isSorted = true;
+
+        for (let i = 1; i < this.length; i++) {
+          previousValue = this[i - 1];
+          currentValue = this[i];
```
BudnikOleksii, line 18:
```suggestion
          const previousValue = this[i - 1];
          const currentValue = this[i];
```
As you don't use these values outside the loop you can create them right here

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, check the comments

src/arrayMethodSort.js
```diff
@@ -4,9 +4,34 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function(compareFunction = defaultCompareFn) {
+    let isSorted;
+
+    if (compareFunction !== undefined) {
```
BudnikOleksii, line 10: This check is redundant as you assign a default value

BudnikOleksii (approved):
Good work 👍

PR #1765, opened 2024-02-27
add test solution

olehmarushchak:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,27 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
```
anastasiiavorobiova, line 7:
```suggestion
  [].__proto__.sort2 = function(compareFunction = (a, b) => String(a) > String(b) ? 1 : -1)) {
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Great work!

PR #1768, opened 2024-02-29
sorted an array

KatsuboDmytro:


src/arrayMethodSort.js
```diff
@@ -5,8 +5,30 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    const compare = compareFunction || toSort;
```
etojeDenys, line 8: instead of this check, just add the default value for comparateFunction argument

etojeDenys (changes requested):
or don't hesitate to ask for some help in the chat

anastasiiavorobiova (approved):
Great work!

PR #1772, opened 2024-03-02
added needed implementation

RostyslavKuzmych:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
...
+  ) {
+    for (let i = 0; i < this.length; i++) {
+      for (let j = i - 1; j >= 0; j--) {
+        if (compareFunction(this[i], this[j]) < 0) {
+          const lesserValue = this[i];
+          const biggerValue = this[j];
+
+          this[i] = biggerValue;
+          this[j] = lesserValue;
+          i--;
```
sarakhmen, line 20: this swap operation could be simplified and you should use conventional names

src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (
+    compareFunction = (first, second) => {
+      return String(first) >= String(second) ? 1 : -1;
+    },
```
sarakhmen, line 10: as for convention we should return 0 if two values are equal

sarakhmen (changes requested):
lgtm!

sarakhmen (approved):
lgtm!

PR #1775, opened 2024-03-05
Solution

Nirvacsh:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,31 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  const DEFAULT_COMPARE = (a, b) => {
```
stassokolov1, line 7:
```suggestion
  const defaultCompareFn = (a, b) => {
```

will be more informative

stassokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #1776, opened 2024-03-05
Solution

Taras2552:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,33 @@
...
+        if (String(a) < String(b)) {
+          return -1;
+        }
+
+        if (String(a) > String(b)) {
+          return 1;
+        }
+
+        return 0;
+      });
```
etojeDenys, line 20: just add it as a default value for customCompareFunction argument

etojeDenys (changes requested):
feel free to ask me about that in the chat

SanyaBratashchuk (approved):

PR #1781, opened 2024-03-09
add solution

rwlp:
#Notes

 -Notice that we use fuctions before definition in code. This is because we can't edit .js code before '// write code here'. Of course is possible to define fuction within applyCustomSort(), But I think the code is clearer without nesting functions
 
 -We use quickSort Algorithm from book "Grokking Algorithms: An illustrated guide for programmers and other curious people"


src/arrayMethodSort.js
```diff
@@ -4,9 +4,66 @@
...
+    } else {
+      sorted.forEach((e, i) => (this[i] = e));
+    }
+
+    return this;
   };
 }
 
+/* This is the quickSort recursive method for sorting */
+function quick(param, funC) {
```
jv-aquino, line 48: you could also improve the naming of a lot of variables here to a more readable code

src/arrayMethodSort.js
```diff
@@ -4,9 +4,66 @@
...
+
+  for (let i = 1; i < param.length; i++) {
+    if (funC(param[i], pivot) < 0) {
+      less.push(param[i]);
+    } else {
+      greater.push(param[i]);
+    }
+  }
+
+  return quick(less, funC).concat(pivot, quick(greater, funC));
```
jv-aquino, line 66: you could use the spread operator here I think, but this looks even better imo

src/arrayMethodSort.js
```diff
@@ -4,9 +4,66 @@
...
+      sorted.forEach((e, i) => (this[i] = e));
+    }
+
+    return this;
   };
 }
 
+/* This is the quickSort recursive method for sorting */
+function quick(param, funC) {
+  if (param.length < 2) {
```
jv-aquino, line 49: nice optimization!

jv-aquino (changes requested):
Wow Rafael... just woowww!! Nice! About algorithm and stuff I didn't find any problem and I liked your code structure a lot! 
I would approve first hand if it had a deadline, but as it is I think you could take some time to make it better in terms of naming conventions (take your time to improve this, it's worth it!)

jv-aquino (approved):
Great job!

PR #1787, opened 2024-03-11
liquidated

true-binar-White-Rabbit:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+            return 1;
+
+          case a < b:
+            return -1;
+
+          default:
+            return 0;
+        }
+      };
+    }
```
etojeDenys, line 25: just add the default value for callback instead of this 
```suggestion
  [].__proto__.sort2 = function (callback = (a,b) => {...}) {
```

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1788, opened 2024-03-12
implementation of the sort method

IvanPovich:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,23 @@
 /**
  * Implement method Sort
  */
+
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (
+    compareFunction = (a, b) => String(a) > String(b),
```
nazarmatsevych, line 9: You can create a separate function and use it like a default value for compareFunction

nazarmatsevych (approved):
Good work, please check the comment below

PR #1792, opened 2024-03-13
task complete

boikoua:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,32 @@
...
+    return 1;
+  }
+
+  if (String(a) < String(b)) {
+    return -1;
+  }
+
+  return 0;
+};
+
```
l4st1m0za, line 34: It's a good thing that you have decided to do like this, but there is one much faster way. You could do like this 
`function (compareFunction = (a, b) => `${a}` > `${b}`) {`

l4st1m0za (approved):
Good job!

PR #1799, opened 2024-03-15
Solution

webdevnikfull:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,8 @@
 'use strict';
 
-/**
- * Implement method Sort
```
DmytroOleksiiuk, line 4: Please write your implementation here, not in another method

DmytroOleksiiuk (changes requested):
Please check my comment. Also the 'sort2' method in the first code snippet lacks enclosing parentheses around the default parameter assignment.

PR #1814, opened 2024-03-21
add solution

Vitalii-Belostotskiy:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,23 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  function defaultCompare(a, b) {
+    return a.toString() > b.toString();
+  }
+
+  [].__proto__.sort2 = function (compareFunction = defaultCompare) {
     // write code here
```
OleziO, line 12: Remove all comments from the code.

OleziO (changes requested):

PR #1820, opened 2024-03-27
implement original sort method

BOHDANLIUYUI:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
...
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
+    let callback = compareFunction;
+
+    if (!compareFunction) {
+      callback = (arg1, arg2) => {
+        return '' + arg1 > '' + arg2 ? 1 : -1;
+      };
```
volodymyr-soltys97, line 13: Move callback logic to the separate function take it outside the current function and pass it as the default function parameter.
Also you need to rewrite this function like this
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better!

lerastarynets (approved):
You're on the right track!

PR #1824, opened 2024-03-28
feat: add solution

mfcastilho:


EdPirro (changes requested):
Awesome job, @mfcastilho! Very clean and straightforward code 👏!

Your current implementation uses a [bubble sort](https://en.wikipedia.org/wiki/Bubble_sort) approach, which is very intuitive, but less efficient. 
As an additional challenge would you be able to implement this sort using a [quick sort](https://en.wikipedia.org/wiki/Quicksort) or a [merge sort](https://en.wikipedia.org/wiki/Merge_sort)? (this would greatly increase your code performance from O(n²) to O(nlogn))

**This is a personal suggestion and not a required change; if you don't wish to take on this challenge just let me know and I'll approve your submission as is**

DarkMistyRoom (approved):
Good job!

PR #1830, opened 2024-04-02
Solution

ArtemZakhar:
I don't like this part of the code:

`        if (
          (compareResult && compareFunction === undefined) ||
          (compareFunction && compareResult <= 0)
        )
`

I was trying to make a separate function and set it as a default value to "compareFunction" but failed(


metrakonon (approved):
:thumbsup:

DanilWeda (approved):
Great! 
Approved.

PR #1834, opened 2024-04-04
Added solution

SlavaRevit:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,37 @@
...
+
+          if (result > 0) {
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
+      }
+    } else if (callback === undefined) {
```
TarasHoliuk, line 18: here you duplicate part of the logic if the callback is undefined

instead, set default value for callback argument
```
function(callback = defaultCallback)
```
define default callback above the current function

TarasHoliuk (changes requested):

vadiimvooo (approved):
Well done!)

PR #1835, opened 2024-04-05
add solution

iAlfreX:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
metrakonon, line 7: You can create compareFn  outside of applyCustomSort() function and set it as default.
It will look like this.
317202373-90408886-606c-4da6-9b68-915c0ea6c6ae

src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+    if (typeof compareFunction === 'function') {
```
metrakonon, line 8: Then you can remove  this if statement .

src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+      for (let i = 0; i < this.length; i++) {
+        for (let j = i + 1; j < this.length; j++) {
+          if (compareFunction(this[i], this[j]) > 0) {
```
metrakonon, line 11: Call here new function witch you write outside this function(like in example above).

src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+        if (this[i].toString() > this[j].toString()) {
+          const temp = this[j];
+
+          this[j] = this[i];
+          this[i] = temp;
+        }
+      }
+    }
+
+    return this;
```
metrakonon, line 34: After all changes, you should be able to remove this entire section, which optimizes your code significantly and reduces the number of loops from 4 to 2.

metrakonon (changes requested):
You did a great job! Here are some improvements you can make to optimize your code.

src/arrayMethodSort.js
iAlfreX: Okay, thanks for clarifying)

iAlfreX:
Done!

metrakonon (approved):
:+1:

PR #1837, opened 2024-04-07
Solution

alexmotalex:
Hi there:) I tried to implement this task using a quick sort, but unfortunately, I did'n figure out how to solve it. 

function qSort(arr) {
  if (arr.length <= 1) {
    return arr;
  }

  const pivot = String(arr[Math.floor(arr.length / 2)]);
  const less = [];
  const greater = [];

  for (let i = 1; i < arr.length; i++) {
    const current = String(arr[i]);
    if (current < pivot) {
      less.push(current);
    } else {
      greater.push(current);
    }
  }

  return [...qSort(less), pivot, ...qSort(greater)];
}

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -124,9 +124,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.23.10",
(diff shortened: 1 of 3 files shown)


etojeDenys (approved):
lgtm

PR #1838, opened 2024-04-08
Solution

yuliaPel:


src/arrayMethodSort.js
```diff
@@ -3,9 +3,31 @@
 /**
  * Implement method Sort
  */
+const compareFunctionDefault = (item1, item2) => {
+  const stringItem1 = typeof item1 === 'string' ? item1 : `${item1}`;
+  const stringItem2 = typeof item2 === 'string' ? item2 : `${item2}`;
```
polosanya, line 8: I think this check is redundant. Just wrap items into String() or `${}` anyway

polosanya (changes requested):
Almost done!

polosanya (approved):
Well done!

PR #1854, opened 2024-04-17
added sort function implementation

MarkGodnyi:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,53 @@
...
+      this[itemIndex] = this[currentEnd - 1];
+      this[currentEnd - 1] = itemToBePlacedAtTheEnd;
+      currentEnd--;
+      currentSearchIndex = 0;
+      itemToBePlacedAtTheEnd = this[0];
+      itemIndex = 0;
+    }
+
+    return this;
   };
```
nazarmatsevych, line 54:
```suggestion
  [].__proto__.sort2 = function (compareFunction = compareDefault) {
    for (let i = 0; i < this.length - 1; i++) {
      for (let j = i + 1; j < this.length; j++) {
        if (compareFunction(this[i], this[j]) > 0) {
          const temp = this[i];

          this[i] = this[j];
          this[j] = temp;
        }
      }
    }
```

Your solution looks too complicated and we can simplify it. Due to sort method documentation we need to swap elements if the callback function returns a value greater than zero

src/arrayMethodSort.js
```diff
@@ -4,8 +4,53 @@
...
+
+    if (areInSameCase) {
+      return aString.localeCompare(b);
+    }
+
+    if (isUpperCase(a)) {
+      return -1;
+    } else {
+      return 1;
+    }
```
nazarmatsevych, line 22:
```suggestion
  if (String(a) > String(b)) {
    return 1;
  }

  if (String(a) < String(b)) {
    return -1;
  }
  
  return 0;
```

Similarly, here we can simplify, we don't need to check whether it is uppercase or not

nazarmatsevych (changes requested):
Good start,  please check the comment below.

If you have any questions - feel free to ask them in fe_chat

vadiimvooo (approved):
Well done!)

PR #1864, opened 2024-04-24
add solution

Opokhvalenko:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,25 @@
...
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
+      }
+
+      return this;
+    }
```
etojeDenys, line 25: we can simplify it


```suggestion
  const defaultSort = [].sort;

  [].__proto__.sort2 = function (compareFunction = ...) { // add here the default compareFunction value
      for (let i = 0; i < this.length - 1; i++) {
        for (let j = 0; j < this.length - i - 1; j++) {
          if (compareFunction(this[j], this[j+1])) { // and just call this function here
            const temp = this[j];

            this[j] = this[j + 1];
            this[j + 1] = temp;
          }
        }
      }

      return this;
    }
```

etojeDenys (changes requested):
feel free to ask for some help in the chat

src/arrayMethodSort.js
```diff
@@ -4,9 +4,32 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  // const defaultSort = [].sort;
```
vadiimvooo, line 7:
```suggestion
```

vadiimvooo (approved):
LGTM

PR #1868, opened 2024-04-29
add solution

BohdanDymydiuk:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,27 @@
...
+      }
+
+      if (compareFunction) {
+        for (let j = i; j < this.length; j++) {
+          if (compareFunction(this[i], this[j]) === 1) {
+            this.splice(i, 0, this[j]);
+            this.splice(j + 1, 1);
+          }
+        }
+      }
```
volodymyr-soltys97, line 25: This logic is now very similar and duplicated, so you need to create a helper function as the documentation says and use it. And provide this function as default parameter to the method
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Great work 🔥

PR #1879, opened 2024-05-08
GG Mate

Teslaanazuma:
Write your own implementation of Array.prototype.sort method. Use predefined [].__proto__.sort2 function for your implementation.


anastasiiavorobiova (approved):
Great work!

PR #1895, opened 2024-05-28
add solution

RuslanHnatyshyn:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  function defaultCompare(a, b) {
+    if (a.toString() < b.toString()) {
+      return -1;
+    }
+
+    if (a.toString() > b.toString()) {
+      return 1;
+    }
```
yuliaPel, line 14: Good job. But I think you should use a variable for a.toString() and b.toString() to avoid code duplication.

yuliaPel (approved):

polosanya (approved):
Good job!

RuslanHnatyshyn:
Thank you for your comments.

PR #1899, opened 2024-06-06
My solution to task

Karanelus:
Write your own implementation of `Array.prototype.sort` method. Use predefined `[].__proto__.sort2` function for your implementation.

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -4,8 +4,46 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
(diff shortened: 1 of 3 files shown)


natalia-klonowska (approved):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,46 @@
...
+        if (aToString < bToString) {
+          return -1;
+        }
+
+        if (aToString > bToString) {
+          return 1;
+        }
+
+        return 0;
+      };
```
natalia-klonowska, line 28: to improve readability you could create as separate function

natalia-klonowska (commented):
Nice work 😃

PR #1904, opened 2024-06-11
add task solution

yursprava:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    if (typeof compareFunction !== 'function') {
+      compareFunction = (a, b) => {
```
lerastarynets, line 9: Assignment to function parameter 'compareFunction' is an error, don't do like so

src/arrayMethodSort.js
```diff
@@ -5,7 +5,38 @@
...
+          return 1;
+        }
+
+        return 0;
+      };
+    }
+
+    const array = this;
+    const length = array.length;
+
```
lerastarynets, line 27: all the above code not needed, you can shorten this to
```
[].__proto__.sort2 = function (
    compareFunction = (a, b) => String(a) > String(b),
  ) {
    // write code here
      for (let index = i + 1; index < this.length; index++) {
        if (compareFunction(this[i], this[index]) > 0) {
          const value = this[i];

          this[i] = this[index];
          this[index] = value;
        }
      }
    }
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

lerastarynets (approved):
You're doing a fantastic job, keep going!

PR #1907, opened 2024-06-22
Solution

bhoriunov:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
VitaliyBondarenko1982, line 7:
```suggestion
  [].__proto__.sort2 = function (compareFunction = defaultCompareFunction) {
```
move defaultCompareFunction outside of sort2 function and use it as default argument.
in method work with compareFunction (no need make compare variable)

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some imrpovements.

src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction = defaultCompareFunction) {
+    const compare = compareFunction || defaultCompareFunction;
```
SanyaBratashchuk, line 8:
```suggestion
```

src/arrayMethodSort.js
```diff
@@ -4,9 +4,35 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      for (let j = 0; j < this.length - i - 1; j++) {
+        if (compare(this[j], this[j + 1]) > 0) {
```
SanyaBratashchuk, line 12:
```suggestion
        if (compareFunction(this[j], this[j + 1]) > 0) {
```

SanyaBratashchuk (approved):

PR #1908, opened 2024-06-24
Solution

olena0112:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,35 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
```suggestion
  [].__proto__.sort2 = function (compareFunction = ...) {
```
just add the default value for compareFunction

src/arrayMethodSort.js
```diff
@@ -4,8 +4,35 @@
...
+
+        if (strA < strB) {
+          return -1;
+        }
+
+        if (strA > strB) {
+          return 1;
+        }
+
+        return 0;
```
etojeDenys, line 22: instead of this logic

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #1912, opened 2024-07-03
Solution v 1.0

StanislavKapytsia:
Write your own implementation of Array.prototype.sort method. Use predefined [].__proto__.sort2 function for your implementation.

Read more about [Array.prototype.sort()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)


volodymyr-soltys97 (approved):
Well done! 🔥

PR #1919, opened 2024-07-10
add solution

RimmaBuzuluk:


src/arrayMethodSort.js
```diff
@@ -2,10 +2,42 @@
...
+  }
+
+  if (aNum > bNum) {
+    return 1;
+  }
+
+  if (aNum === bNum) {
+    return 0;
+  }
+};
```
Teollan, line 24: This function does the job, but the name `func` is not informative. 

Consider renaming it to `defaultCompare` or something like that.

Teollan (changes requested):
Great job :fire:  !

Please, fix one minor issue, to make it perfect!

DanilWeda (approved):
Great job! 
Approved. 
Many thanks)

PR #1924, opened 2024-07-11
Solution

vtroni:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
etojeDenys, line 7: instead of that if statement let's just specify the default value
```suggestion
  [].__proto__.sort2 = function (compareFunction = ........) {
```

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1944, opened 2024-07-28
add solution

George-Buzhenitsa:
I checked if there was a callback function given, and if it wasn't I just sorted elments with a bubble sort algorithm, otherwise I called the function from callback and just checked if the return of the callback function is positive)

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -4,8 +4,36 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
+    if (compareFunction === undefined) {
+      for (let i = 0; i < this.length; i++) {
+        for (let k = 0; k < this.length; k++) {
+          if ('' + this[k] > '' + this[k + 1]) {
+            const temp = this[k];
+
+            this[k] = this[k + 1];
+            this[k + 1] = temp;
+          }
+        }
+      }
+
+      return this;
+    }
+
+    if (compareFunction) {
+      for (let i = 0; i < this.length; i++) {
+        for (let k = 0; k < this.length; k++) {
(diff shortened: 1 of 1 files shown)


anastasiiavorobiova (changes requested):
Great start! Consider fixing AI comments

anastasiiavorobiova (approved):
Awesome!

src/arrayMethodSort.js
```diff
@@ -4,8 +4,36 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
```
AvinashMarkad, line 8: /**
 * Bubble sort implementation with optional callback function.
 * @param {number[]} arr - The array to be sorted.
 * @param {function} [callback] - Optional. A function that defines the sort order.
 * @return {number[]} - The sorted array.
 */
function bubbleSort(arr, callback) {
    let n = arr.length;
    let swapped;
    for (let i = 0; i < n; i++) {
        swapped = false;
        for (let j = 0; j < n - 1 - i; j++) {
            let comparison;
            if (callback) {
                comparison = callback(arr[j], arr[j + 1]);
            } else {
                comparison = arr[j] - arr[j + 1];
            }

            if (comparison > 0) {
                // Swap arr[j] and arr[j + 1]
                [arr[j], arr[j + 1]] = [arr[j + 1], arr[j]];
                swapped = true;
            }
        }

        // If no elements were swapped, the array is already sorted
        if (!swapped) break;
    }
    return arr;
}

// Usage without callback (default ascending order)
console.log(bubbleSort([4, 2, 7, 1, 3])); // Output: [1, 2, 3, 4, 7]

// Usage with a callback (descending order)
console.log(bubbleSort([4, 2, 7, 1, 3], (a, b) => b - a)); // Output: [7, 4, 3, 2, 1]

AvinashMarkad (changes requested):

PR #1945, opened 2024-07-30
Solution

jacube1:


Zibi95 (approved):
Well done. Can you tell what kind of sort method is that? 😁

jacube1:
Bubble sort :)

PR #1970, opened 2024-08-29
solved task

PikalovAndrey:


src/arrayMethodSort.js
```diff
@@ -1,11 +1,32 @@
...
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
+    if (typeof compareFunction !== 'function') {
+      compareFunction = function (a, b) {
+        const strA = String(a);
+        const strB = String(b);
+
+        return strA < strB ? -1 : strA > strB ? 1 : 0;
+      };
+    }
```
BudnikOleksii, line 16: Move the function outside this function and use the default parameters to set this function if it's not passed https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

src/arrayMethodSort.js
```diff
@@ -1,11 +1,32 @@
...
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
+    if (typeof compareFunction !== 'function') {
+      compareFunction = function (a, b) {
+        const strA = String(a);
+        const strB = String(b);
+
+        return strA < strB ? -1 : strA > strB ? 1 : 0;
```
BudnikOleksii, line 14: Avoid nested ternary operators

src/arrayMethodSort.js
```diff
@@ -1,11 +1,32 @@
+/* eslint-disable no-param-reassign */
```
BudnikOleksii, line 1: Don't disable linter

BudnikOleksii (changes requested):
Good work 👍 
Let's fix a few moments

src/arrayMethodSort.js
PikalovAndrey: OK, but why? It's not so hard to understand the logic behind this strip of code.

src/arrayMethodSort.js
```diff
@@ -3,9 +3,39 @@
...
+  const strA = String(a);
+  const strB = String(b);
+
+  if (strA < strB) {
+    return -1;
+  } else if (strA > strB) {
+    return 1;
+  } else {
+    return 0;
+  }
```
Moroz-Dmytro, line 17:
```suggestion
  if (strA < strB) {
    return -1;
  }
  
  if (strA > strB) {
    return 1;
  }
  
  return 0;
```

Moroz-Dmytro (approved):
Good work 👍

PR #1976, opened 2024-09-01
Solution

luchali:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  [].__proto__.sort2 = function (compareFunction) {
```
```suggestion
  [].__proto__.sort2 = function (compareFunction = ......) {
```
instead of all those checks, just specify the default value for the argument

etojeDenys (changes requested):

lerastarynets (approved):
Keep up the great work!

PR #1979, opened 2024-09-02
test

OleksandrFEdev:


src/arrayMethodSort.js
```diff
@@ -4,9 +4,33 @@
...
+      for (let i = 0; i < n - 1; i++) {
+        if (compareFunction) {
+          if (compareFunction(this[i], this[i + 1]) > 0) {
+            [this[i], this[i + 1]] = [this[i + 1], this[i]];
+            swapped = true;
+          }
+        } else {
+          if (String(this[i]) > String(this[i + 1])) {
+            [this[i], this[i + 1]] = [this[i + 1], this[i]];
+            swapped = true;
```
TarasHoliuk, line 23: You repeat this twice

Instead define default compare function above and use it as a default value for compareFunction variable

TarasHoliuk (changes requested):
Great job 👍 
One comment added

lerastarynets (approved):
lgtm 🚀

PR #1985, opened 2024-09-04
Solution

alena-bondar:


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix code from AI comments
Also, implement the compare function 
image

anastasiiavorobiova (approved):
Great work!

PR #1986, opened 2024-09-04
Solution

Valeraom:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
...
+            const firstElement = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = firstElement;
+          }
+        }
+      }
+    }
+
+    return this;
```
VolodymyrKirichenko: 1) u are duplicating code.
make an additional function that will run when compareFunction is not given to u
2) Also, I think u should use a `do...while` loop to keep the loop running until the entire array is sorted
```suggestion
  [].__proto__.sort2 = function (compareFunction = additionalCompareFunction) {
    for (let i = 0; i < this.length - 1; i++) {
      for (let j = 0; j < this.length - 1; j++) {
        if (compareFunction(String(this[j]), String(this[j + 1])) > 0) {
          const firstElement = this[j];

          this[j] = this[j + 1];
          this[j + 1] = firstElement;
        }
      }
    }
    
    function additionalCompareFunction = (a, b) {
      ........
    }
```

VolodymyrKirichenko (changes requested):
Overall - good job, but let's improve ur code! 👍

src/arrayMethodSort.js
```diff
@@ -4,9 +4,32 @@
...
+          isChanged = true;
+        }
+      }
+    } while (isChanged);
+
+    return this;
   };
 }
 
+function additionalCompareFunction(a, b) {
```
VolodymyrKirichenko, line 29: The comparison function(`additionalCompareFunction`) for the sort method must return one of three values: 1, 0, or -1. This is required for correct comparison of array elements

```
if (String(a) > String(b)) {
  return 1;
}
  
if (===) {
}
  
if (<) {
}
```

VolodymyrKirichenko (approved):
Well done, but review my comments and fix them 🔥

PR #1990, opened 2024-09-09
add solution

Tulenty:


src/arrayMethodSort.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomSort() {
   [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+    compareFunction = compareFunction || function(a,b){
```
etojeDenys: instead of all those checks, just add the default value for the compareFunction parameter

etojeDenys (changes requested):

etojeDenys (approved):
🔥

PR #2006, opened 2024-09-27
Solution

KrivorotkoYurii:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,37 @@
...
+        if (string1 > string2) {
+          return 1;
+        }
+
+        if (string1 < string2) {
+          return -1;
+        }
+
+        return 0;
+      };
```
BudnikOleksii: Move default function outside

src/arrayMethodSort.js
```diff
@@ -4,8 +4,37 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
BudnikOleksii:
```suggestion
  [].__proto__.sort2 = function (compareFunction = defaultCompareFunction) {
```
You can use default parameters https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

BudnikOleksii (approved):
Good job 👍

PR #2010, opened 2024-09-30
solution

minimal005:
Write your own implementation of `Array.prototype.sort` method. Use predefined `[].__proto__.sort2` function for your implementation.

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -1,11 +1,17 @@
 'use strict';
 
-/**
+/*
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction = undefined) {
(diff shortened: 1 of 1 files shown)


src/arrayMethodSort.js
```diff
@@ -1,11 +1,25 @@
...
-/**
+/*
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction = undefined) {
+    const myOwnCompare =
+      compareFunction || ((a, b) => (a.toString() > b.toString() ? 1 : -1));
```
BudnikOleksii, line 9:
```suggestion
  [].__proto__.sort2 = function (compareFunction = defaultCompareFunction) {
```
- You can move outside the function and as a default parameter
- Name is strange

BudnikOleksii (approved):
Good work 👍 
Left a suggestion, check it

PR #2043, opened 2024-10-29
add task solution

Ga1dar:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
+  [].__proto__.sort2 = function (compareFunction) {
```
etojeDenys: just specify the default value
```suggestion
  [].__proto__.sort2 = function (compareFunction = (a, b) => String(a) > String(b) ? 1 : -1) {
```

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+        const diff = strA.charCodeAt(i) - strB.charCodeAt(i);
+
+        if (diff !== 0) {
+          return diff;
+        }
+      }
+
+      return strA.length - strB.length;
+    };
+
```
etojeDenys: this is redundant

src/arrayMethodSort.js
```diff
@@ -4,8 +4,38 @@
...
+    const d = compareFunction || defaultCompare;
```
etojeDenys: and this

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Well done

PR #2052, opened 2024-11-04
Solution

iirk1:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,26 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
etojeDenys:
```suggestion
  [].__proto__.sort2 = function (compareFunction = ....) {
```
just specify the default value for this function instead of some checks

etojeDenys (changes requested):

volodymyr-soltys97 (approved):
Great work! 🔥

PR #2056, opened 2024-11-05
solution

k0tlik02:
Used bubble sort for sorting array

### DIFF
--- a/src/arrayMethodSort.js
+++ b/src/arrayMethodSort.js
@@ -4,8 +4,27 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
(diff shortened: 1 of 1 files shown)


src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
BudnikOleksii, line 7:
```suggestion
  [].__proto__.sort2 = function (compareFunction = defaultCompareFunction) {
```
It's better to assign default function as a default parameter, in this case you can get rid of if/else

BudnikOleksii (approved):
Good job 👍

PR #2061, opened 2024-11-12
Solution

schyrva:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,35 @@
...
+        if (strA < strB) {
+          return -1;
+        }
+
+        if (strA > strB) {
+          return 1;
+        }
+
+        return 0;
+      };
```
BudnikOleksii: You can move the default function outside the method and set it as a default parameter

BudnikOleksii (approved):
Good work 👍

PR #2078, opened 2024-12-02
Result

Glasis9:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,34 @@
...
+          return -1;
+        }
+
+        if (a.toString() > b.toString()) {
+          return 1;
+        }
+
+        return 0;
+      };
+    }
```
etojeDenys: just specify the default value for the argument

```suggestion
  [].__proto__.sort2 = function (compareFunction = ...) {
    }
```

etojeDenys (changes requested):

src/arrayMethodSort.js
```diff
@@ -4,8 +4,31 @@
...
+        return -1;
+      }
+
+      if (a.toString() > b.toString()) {
+        return 1;
+      }
+
+      return 0;
+    },
+  ) {
```
volodymyr-soltys97, line 19: You can simplify like this
```suggestion
  [].__proto__.sort2 = function (compareFunction = (a, b) => String(a) > String(b) ? 1 : -1)) {
```

volodymyr-soltys97 (approved):
Good job 👍

PR #2082, opened 2024-12-05
Solution

IvanVaniuk:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
...
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
+    let compFunc = compareFunction;
+
+    if (typeof compFunc !== 'function') {
+      compFunc = (a, b) => {
+        return String(a) === String(b) ? 0 : String(a) > String(b) ? 1 : -1;
+      };
+    }
```
etojeDenys: you don't need this check

src/arrayMethodSort.js
```diff
@@ -4,8 +4,27 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
etojeDenys, line 7: just add the default value here
```suggestion
  [].__proto__.sort2 = function (compareFunction = ...) {
```

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Looks good

PR #2101, opened 2024-12-25
add task solution

Uusuff:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,39 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
etojeDenys:
```suggestion
  [].__proto__.sort2 = function (compareFunction = .....) {
```

just provide here the default value

etojeDenys (changes requested):

etojeDenys (approved):

PR #2149, opened 2025-02-19
Solution

fromjkeee16:


alexander-ignatow (approved):
Good job 👍 , but the built-in sort is more effective :)

PR #2244, opened 2025-05-05
Solution

Ynln69:


lidiia-tsymborovych (approved):
Good job! It just needs a little simplification.

PR #2253, opened 2025-05-15
Solution 15.05.25

Alya-ya:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,28 @@
...
+            this[j + 1] = temp;
+          }
+        } else {
+          if (String(this[j]) > String(this[j + 1])) {
+            const temp = this[j];
+
+            this[j] = this[j + 1];
+            this[j + 1] = temp;
+          }
+        }
```
volodymyr-soltys97: This logic is practically the same, you don't need to duplicate it, instead you can pass a default value for the parameter `compareFunction`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (changes requested):
Hi, here is no changes from the previous review

anastasiiavorobiova (approved):
Great work!

PR #2324, opened 2025-08-03
add solution

derpedcatto:


Yurii-Kupchynskyi (approved):
The swapNums function works correctly. The logic is simple and clear.

PR #2356, opened 2025-09-09
_create sort method_

S-Khraban:


igorkruz (changes requested):
Add the demo link to your project, because without it I can’t review it.

src/arrayMethodSort.js
```diff
@@ -1,11 +1,36 @@
 'use strict';
 
-/**
- * Implement method Sort
- */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
```
igorkruz, line 4: Avoid mutating __proto__, Instead, extend the prototype directly:
`Array.prototype.sort2 = function (compareFunction) {`

src/arrayMethodSort.js
```diff
@@ -1,11 +1,36 @@
...
+        if (A < B) {
+          return -1;
+        }
+
+        if (A > B) {
+          return 1;
+        }
+
+        return 0;
+      };
```
igorkruz, line 20: Wrapping everything with String() is unnecessary. You can just compare directly:

if (a < b) return -1;
if (a > b) return 1;
return 0;

src/arrayMethodSort.js
```diff
@@ -1,11 +1,36 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        if (cmp(this[j], this[j + 1]) > 0) {
+          const temp = this[j];
+
+          this[j] = this[j + 1];
+          this[j + 1] = temp;
```
igorkruz, line 28: Swapping values, сan be simplified using destructuring: 
`[this[j], this[j + 1]] = [this[j + 1], this[j]];`

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

PR #2378, opened 2025-10-04
Solution

pzwolenik:


pzwolenik:
Addition to critical issueas: Could not override Array.prototype.sort due to environment restrictions.

danon321 (approved):

PR #2420, opened 2025-12-15
Bubble sort

Dimononon:


src/arrayMethodSort.js
```diff
@@ -4,8 +4,35 @@
...
+
+        if (strA > strB) {
+          return 1;
+        }
+
+        return 0;
+      };
+    }
+
+    for (let i = 0; i < this.length; i++) {
```
dmytro-halieba, line 27: The algorithm always runs in O(n²), even if the array is already sorted.
This can be improved by adding a swapped flag and breaking out of the loop early when no swaps occur, which indicates that the array is already sorted.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,35 @@
...
+      for (let j = 0; j < this.length - 1 - i; j++) {
+        if (compareFunctionLocal(this[j], this[j + 1]) > 0) {
+          [this[j], this[j + 1]] = [this[j + 1], this[j]];
```
dmytro-halieba: Using array destructuring for swapping creates a temporary array on each swap.
This adds unnecessary memory allocations and overhead. Using a simple temporary variable (temp) is more memory-efficient and generally faster in tight loops.

src/arrayMethodSort.js
```diff
@@ -4,8 +4,35 @@
  * Implement method Sort
  */
 function applyCustomSort() {
-  [].__proto__.sort2 = function(compareFunction) {
-    // write code here
+  [].__proto__.sort2 = function (compareFunction) {
+    let compareFunctionLocal = compareFunction;
+
+    if (compareFunction === undefined) {
```
dmytro-halieba: The check `compareFunction === undefined` is not sufficient.
The implementation does not handle invalid arguments (e.g. numbers or strings passed as compareFunction). It would be safer to explicitly verify that the argument is a function and provide a fallback or throw an error otherwise.

dmytro-halieba (changes requested):
Overall, this is a solid and correct implementation of a custom sort method.
The solution follows the expected behavior, handles the default comparison logic properly, and demonstrates a good understanding of sorting fundamentals.

However, there are a few important improvement areas:

- The algorithm can be optimized to avoid unnecessary iterations when the array is already sorted.

- The swap implementation could be made more efficient to reduce memory overhead.

- Input validation for the compareFunction argument is missing, which may lead to unexpected behavior when invalid values are passed.

**The only required change is to properly handle invalid inputs passed as compareFunction.**
_Hint: you can use typeof to validate that it is a function._
All other comments are optional improvements and do not affect the correctness of the implementation.

PR #2448, opened 2026-01-20
Develop

TarasPidkuimukha:


Denys-Kravchuk9988 (changes requested):
Good job! 
But I would recommend to apply all comments from AI. 

1. Sort function should return numbers (1, 0, -1)
2. Performance fix

